PR #32, opened 2014-06-27
Micro time ago

dgraham:
Adds a micro formatter that shortens descriptions to a single letter. Useful for lists of recently received items, like tweets and emails.

``` html
<time is="time-ago" format="micro" datetime="2014-06-27T17:56:00Z">
  43m
</time>
```

### DIFF
--- a/time-elements.js
+++ b/time-elements.js
@@ -84,70 +84,16 @@
     });
   }
 
-
-  function CalendarDate(year, month, day) {
-    this.date = new Date(Date.UTC(year, month - 1));
-    this.date.setUTCDate(day);
-    this.year = this.date.getUTCFullYear();
-    this.month = this.date.getUTCMonth() + 1;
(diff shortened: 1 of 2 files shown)


time-elements.js
```diff
@@ -163,6 +163,27 @@
...
+    var sec = ms / 1000;
+    var min = sec / 60;
+    var hr = min / 60;
+    var day = hr / 24;
+    var month = day / 30;
+    var year = month / 12;
+    if (min < 1) {
+      return '1m';
+    } else if (min < 60) {
+      return Math.round(min) + 'm';
```
josh: @dgraham do you remember why this rounds after comparison, not before? The other non-micro checks round out all the values first.

dgraham: I don't remember why they round differently. I just ran the tests locally, after moving the `round` calls before the comparison, and everything passes. Should be fine either way.

PR #38, opened 2014-09-09
Future times with realtive-time-to-go and time-to-go elements

BradDenver:
The use case for this commit is when you wish to display time relative to expected future dates such as in the case of expiry dates.

This is implemented via to new elements, realtive-time-to-go and time-to-go.

This commit also make the 'just now' string configurable via a just-now attribute.


mislav:
There is future time support now with `time-until` elements.

BradDenver:
thanks for the update @mislav

PR #42, opened 2014-10-25
Allow relative-time to handle future dates

markmcspadden:
Why are we soooo concerned with the past? The future is where it's at!!! :rocket: :dizzy: 

We are looking at starting to use web components for travel applications and many of the dates and times associated are in the future. (I did find one lone place for future dates on GitHub, the milestone due date...so it at least has _some_ GH application.)

This PR adds relative-time results that are based off of the same time-ago intervals, just oriented towards the future. "3 days from now", "1 year from now", etc. It also adds a "time-until" option to counter "time-ago" in which the browser will ALWAYS give a future date.

There are several structural changes I made in able to do this so please take a look and provide any feedback.  

Open to any feedback, especially on the wordage...I went back and forth between "3 days from now" and "in 3 days"

### DIFF
--- a/time-elements.js
+++ b/time-elements.js
@@ -202,30 +202,28 @@
     var day = Math.round(hr / 24);
     var month = Math.round(day / 30);
     var year = Math.round(month / 12);
-    if (month > 18) {
+    if (month >= 18) {
       return year + ' years from now';
-    } else if (month > 12) {
+    } else if (month >= 12) {
       return 'a year from now';
-    } else if (month > 2) {
+    } else if (day >= 45) {
       return month + ' months from now';
-    } else if (day > 45) {
+    } else if (day >= 30) {
       return 'a month from now';
-    } else if (day > 1) {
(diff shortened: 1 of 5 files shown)


time-elements.js
```diff
@@ -184,6 +192,66 @@
...
+    } else if (min >= 90) {
+      return hr + ' hours from now';
+    } else if (min >= 45) {
+      return 'an hour from now';
+    } else if (sec >= 90) {
+      return min + ' minutes from now';
+    } else if (sec >= 45) {
+      return 'a minute from now';
+    } else if (sec >= 10) {
+      return sec + ' seconds from now';
```
dgraham: This range is sensitive to a user's browser clock skew not matching the server clock that generated the element. For example, when you submit a comment via a form, and the comment partial is returned from the server, you may see that you wrote your comment "11 seconds from now".

I think we should treat anything under the 90 second range as the "just now" label to give us some buffer between clock differences.

PR #49, opened 2015-04-10
Prepare for npm publish: add package.json fields

wbinnssmith:
This removes the `"private": true` field and adds a main field so time elements can be used with just `require('time-elements')`. 

I'd love to see time-elements on npm :smile: 

Related but not coupled to this: CommonJS (or others for that matter) - the module currently doesn't export anything, since everything is mostly side-effect driven anyway. Maybe it could export the three constructors? Wondering what you think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,7 @@
 {
-  "private": true,
+  "name": "time-elements",
+  "version": "0.4.0",
+  "main": "time-elements.js",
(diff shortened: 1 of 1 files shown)


mislav:
Sorry we haven't seen this! This was done in d522a8debd9facdeef5b0b3e9c04c00559e7710b.

PR #51, opened 2015-05-11
Adds "web-components" keyword

zenorocha:
Which allows customelements.io to fetch this package from Bower


josh:
Killed bower support. Is this is valid as an npm keyword?

zenorocha:
Yep

PR #52, opened 2015-05-27
Typo in localized time example

olegstepura:


olegstepura:
This pull request may be wrong if timezone magic comes into play

mislav:
Yes. It's not a typo, but purposefully different time to indicate that the rendered string is in the local timezone rather than a static one.

PR #61, opened 2016-04-15
Avoid non-standard is=""

### LINKED ISSUE #60: Avoid is="" API
Safari 10 will be supporting Custom Elements soon, but won't support `is=""`.

We should rewrite this extension without extending `<time>` and just do `<relative-time>` and `<local-time>`.

Hopefully we can get github.com updated long before so future Enterprise versions will be ready with this change. Unsure if the Polymer Custom Element polyfill will correctly cover a partial Custom Elements implementation.

-
To: @dgraham @mislav 
CC: @notwaldorf @github/js

josh:
Backwards incompatible change.

Closes #60.

-
CC: @mislav @dgraham


time-elements.js
```diff
@@ -500,26 +495,23 @@
...
   // Public: LocalTimeElement constructor.
   //
   //   var time = new LocalTimeElement()
-  //   # => <time is='local-time'></time>
+  //   # => <local-time></local-time>
   //
   window.LocalTimeElement = document.registerElement('local-time', {
-    prototype: LocalTimePrototype,
-    'extends': 'time'
+    prototype: LocalTimePrototype
```
josh: The change to the library is pretty small itself.

mislav:
👍

PR #62, opened 2016-07-01
Future time redux

holman:
This is a continuation of @markmcspadden's excellent pull from a year and a half ago: #42 (which is also related to #38).

Basically, time-elements should support future dates too. I'm living in the future so the present is my past, time-elements is a present, merge if tests pass (they do).

The majority of this is pulled from @markmcspadden's work; I just updated the syntax from the deprecated `<time is="local-time">` form and updated the README.

This pull adds an additional `time-until` element (which we can axe if it's not worthwhile), and it updates `relative-time` to see the future, too.

/cc @josh @dgraham @github/techno

### DIFF
--- a/test/time-until.js
+++ b/test/time-until.js
@@ -1,0 +1,71 @@
+module('time-until');
+
+test('always uses relative dates', function() {
+  var now = new Date(Date.now() + 10 * 365 * 24 * 60 * 60 * 1000).toISOString();
+  var time = document.createElement('time', 'time-until');
+  time.setAttribute('datetime', now);
+  equal(time.textContent, '10 years from now');
+});
+
+test('rewrites from now future datetime to minutes ago', function() {
+  var now = new Date(Date.now() + 3 * 60 * 1000).toISOString();
+  var time = document.createElement('time', 'time-until');
+  time.setAttribute('datetime', now);
+  equal(time.textContent, '3 minutes from now');
+});
+
+test('rewrites a few seconds from now to just now', function() {
+  var now = new Date().toISOString();
+  var time = document.createElement('time', 'time-until');
+  time.setAttribute('datetime', now);
+  equal(time.textContent, 'just now');
+});
+
+test('displays past times as just now', function() {
+  var now = new Date(Date.now() + 3 * 1000).toISOString();
+  var time = document.createElement('time', 'time-until');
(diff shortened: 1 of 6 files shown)


test/time-until.js
```diff
@@ -0,0 +1,71 @@
...
+  time.setAttribute('format', 'micro');
+  equal(time.textContent, '1m');
+});
+
+test('micro formats hours', function() {
+  var now = new Date(Date.now() + 60 * 60 * 1000).toISOString();
+  var time = document.createElement('time-until');
+  time.setAttribute('datetime', now);
+  time.setAttribute('format', 'micro');
+  equal(time.textContent, '1h');
```
josh: Hmm, getting a failure here, `"60m"` instead of a `"1h"`.

time-elements.js
```diff
@@ -184,6 +192,66 @@
...
+  };
+
+  RelativeTime.prototype.microTimeUntil = function() {
+    var ms = this.date.getTime() - (new Date().getTime());
+    var sec = ms / 1000;
+    var min = sec / 60;
+    var hr = min / 60;
+    var day = hr / 24;
+    var month = day / 30;
+    var year = month / 12;
```
josh: I think we might need `Math.round()` wrapped around all these calculations. I think that will fix the `"60m"` test failure I'm seeing.

test/time-until.js
holman: I had that too when I first ran tests, but then after re-running it went away. Didn't look into it past that point.

time-elements.js
josh: Hmm, maybe thats wrong, only the other `ms` helpers do that. Not sure yet.

josh: Maybe @dgraham remembers why the micro format works different https://github.com/github/time-elements/pull/32/files#r69352974

josh:
PR to PR suggestion https://github.com/holman/time-elements/pull/1

holman:
> PR to PR suggestion holman#1

It's a PR to my PR which is based on another PR. That's a new Personal Record for me.

markmcspadden:
Thanks @holman for resurrecting this one. Enjoy the future!

dgraham:
`Auto-merged branch holman-future-time into master with Monster reference`

PR #63, opened 2016-07-05
Encapsulate all DateTimeFormat calls to work around Chrome bug

mislav:
Chrome doesn't deal well with [new timezones added to Windows](https://blogs.technet.microsoft.com/dst2007/2016/06/02/introducing-new-windows-time-zones/).

Ocassionally when `new Intl.DateTimeFormat()` is called without being given an explicit timezone, it will error out with:

    RangeError: Unsupported time zone specified undefined

Since there's nothing we can do about this right now, simply ignore the error and hope that Chrome is going to fix this in the future.

/cc @dgraham

### DIFF
--- a/time-elements.js
+++ b/time-elements.js
@@ -391,7 +391,15 @@
     if (formatter) {
       return formatter.format(this._date);
     } else {
-      return this._date.toLocaleString();
+      try {
+        return this._date.toLocaleString();
+      } catch(e) {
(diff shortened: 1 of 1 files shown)


josh:
Wow! Finally figured this one out.

dgraham:
Nice workaround for this bug. :shipit:

PR #64, opened 2016-07-26
Fix typo in README example

bjeanes:


jerone:
Same as https://github.com/github/time-elements/pull/52

bjeanes:
If the difference is intentional in order to communicate adjustments made for timezone, might I suggest making it _obviously_ different, such as `21:30`?

PR #65, opened 2016-07-26
Add timezone format to <local-time>

stoe:
To be able to additionally display a timezone string with the `<local-time>` element.

### DIFF
--- a/test/local-time.js
+++ b/test/local-time.js
@@ -46,16 +46,6 @@ test('getFormattedDate with only time attributes', function() {
   ok(browser || phantom);
 });
 
-test('getFormattedDate with only timezone attributes', function() {
-  var time = document.createElement('local-time');
-  time.setAttribute('datetime', '1970-01-01T00:00:00.000Z');
-  time.setAttribute('timezone', 'short');
-
-  var value = time.getFormattedDate();
-  var acceptable = ['CEST', 'CET', 'PDT', 'PST'];
-  ok(acceptable.indexOf(value) !== -1);
-});
-
 test('ignores contents if datetime attribute is missing', function() {
   var time = document.createElement('local-time');
   time.setAttribute('year', 'numeric');
--- a/test/local-time.js
+++ b/test/local-time.js
@@ -52,7 +52,7 @@ test('getFormattedDate with only timezone attributes', function() {
   time.setAttribute('timezone', 'short');
 
   var value = time.getFormattedDate();
-  var acceptable = ['CEST', 'CET', 'PDT', 'PST'];
+  var acceptable = ['CEST', 'CET', 'PDT', 'PST', 'GMT', 'UTC'];
   ok(acceptable.indexOf(value) !== -1);
 });
 
--- a/test/local-time.js
+++ b/test/local-time.js
@@ -52,7 +52,7 @@ test('getFormattedDate with only timezone attributes', function() {
   time.setAttribute('timezone', 'short');
 
   var value = time.getFormattedDate();
-  var acceptable = ['CEST', 'CET', 'PDT', 'PST', 'GMT', 'UTC'];
+  var acceptable = ['CEST', 'CET', 'PDT', 'PST'];
   ok(acceptable.indexOf(value) !== -1);
 });
 
--- a/test/local-time.js
+++ b/test/local-time.js
@@ -46,6 +46,16 @@ test('getFormattedDate with only time attributes', function() {
   ok(browser || phantom);
 });
 
+test('getFormattedDate with only timezone attributes', function() {
+  var time = document.createElement('local-time');
+  time.setAttribute('datetime', '1970-01-01T00:00:00.000Z');
+  time.setAttribute('timezone', 'short');
+
+  var value = time.getFormattedDate();
+  var acceptable = ['CEST', 'CET', 'PDT', 'PST'];
+  ok(acceptable.indexOf(value) !== -1);
+});
+
 test('ignores contents if datetime attribute is missing', function() {
   var time = document.createElement('local-time');
   time.setAttribute('year', 'numeric');
--- a/test/local-time.js
+++ b/test/local-time.js
@@ -46,6 +46,36 @@ test('getFormattedDate with only time attributes', function() {
   ok(browser || phantom);
 });
 
+test('getFormattedDate with short timezone attributes', function() {
+  var time = document.createElement('local-time');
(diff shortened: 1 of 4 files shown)


time-elements.js
```diff
@@ -587,6 +589,33 @@
...
+
+    // no timezone format attributes provided
+    if (Object.keys(options).length === 0) {
+      return;
+    }
+
+    var match = el._date.toString().match(/\(([^)]+)\)/);
+    return match ? match[1] : '';
+  }
+
```
dgraham: The [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat) supports a `timeZoneName` option, so we would need to expose that with a `timeZoneName` attribute on the custom element and [pass it into the formatter](https://github.com/github/time-elements/blob/da6eb9f6cfbdb43bb45ef8ff23e2515667ae1214/time-elements.js#L564). That will avoid regular expression parsing of unknown `toString()` text.

test/local-time.js
```diff
@@ -46,6 +46,16 @@ test('getFormattedDate with only time attributes', function() {
   ok(browser || phantom);
 });
 
+test('getFormattedDate with only timezone attributes', function() {
+  var time = document.createElement('local-time');
+  time.setAttribute('datetime', '1970-01-01T00:00:00.000Z');
+  time.setAttribute('timezone', 'short');
+
+  var value = time.getFormattedDate();
+  var acceptable = ['CEST', 'CET', 'PDT', 'PST', 'GMT', 'UTC'];
```
mislav: What if the test suite is ran from a timezone other than the ones listed here?

time-elements.js
```diff
@@ -587,6 +589,33 @@
...
+  function formatTimezone(el) {
+    // retrieve format settings from attributes
+    var options = {
+      timezone: el.getAttribute('timezone')
+    };
+
+    // remove unset format attributes
+    for (var opt in options) {
+      if (!options[opt]) {
+        delete options[opt];
```
mislav: This whole `options` hash dance could be simply replaced with:

```js
var timezoneSetting = el.getAttribute('timezone')
if (!timezoneSetting) return
var tzDisplay = ...
return tzDisplay
```

No `options` hash, looping over its keys, `Object.keys()` check. There is always ever one key (for now).

time-elements.js
mislav: Note that Chrome in a Windows environment can intermittently throw exceptions on `new Intl.DateTimeFormat`. If you wish to use it, always wrap it in a `try..catch` block:

```js
try {
  var format = new Intl.DateTimeFormat()
} catch (e) {
  if (!(e instanceof RangeError)) {
    throw e
  }
}
```

If there was an exception, the timezone information is unavailable, so you should simply not render any timezone string.

stoe:
@dgraham @mislav refactored to use [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat) and also got rid of extra `formatTimezone()`.

Now it just allows to set two additional attributes on the `<local-time>` element, `timeZone` and `timeZoneName`.

This results for example in the following HTML:

```html
<!-- 04:00 when viewed from PST -->
<!-- 13:00 when viewed from CEST -->
<!-- 16:30 when viewed from IST -->
<local-time
  datetime="2016-07-28T04:00-07:00"
  hour="numeric"
  minute="numeric"
  timezone="PST"
  timezonename="short"
  title="Jul 28, 2016, 1:00 PM GMT+2">
13:00
</local-time>
```

time-elements.js
```diff
@@ -561,7 +563,9 @@
     var options = {
       hour: el.getAttribute('hour'),
       minute: el.getAttribute('minute'),
-      second: el.getAttribute('second')
+      second: el.getAttribute('second'),
+      timeZone: el.getAttribute('timeZone'),
+      timeZoneName: el.getAttribute('timeZoneName')
```
mislav: I don't get the purpose of these new attributes. Why would you set a `timezone=PST` value on the server? Wasn't the point of this pull requests to simply opt into showing the local timezone information to the current user, so that they are certain that the timestamp they're seeing is in their local timezone?

stoe: @mislav, the purpose for setting e.g. `timezone=PST` is to explicitly be able to force the timezone used. It defaults to UTC.

> [The time zone to use. The only value implementations must recognize is "UTC"; the default is the runtime's default time zone. Implementations may also recognize the time zone names of the IANA time zone database, such as "Asia/Shanghai", "Asia/Kolkata", "America/New_York".
](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat)

I agree that it is not necessary, but if adding `timeZoneName` I thought to also add `timeZone` for [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat).

mislav: But, if we wanted to force the timezone used, we could just have the server render the timestamp in the specified timezone itself. We wouldn't need JavaScript for that. The purpose of this project is to take a machine-readable timestamp rendered from the server and transform it into a local time based on the user's browser.

stoe: > But, if we wanted to force the timezone used, we could just have the server render the timestamp in the specified timezone itself. We wouldn't need JavaScript for that.

@mislav, you are :100: correct there.

> to force the timezone used

Reading that again, I have not worded that correctly. My understanding is that it doesn't force the rendering of that timezone, but tells JavaScript additionally to the timezone string in `datetime` what the used timezone  is. A little redundant, yes.

You think we should take it out again?

mislav: > You think we should take it out again?

Yes, I think we should focus solely on the original purpose of this PR: making it explicit to the user which timezone the timestamp is rendered in (always theirs). Proposal:

```html
<local-time> → 8 Apr 2018 21:30
<local-time timezone="short"> → 8 Apr 2018 21:30 CEST
<local-time timezone="numeric"> → 8 Apr 2018 21:30 +02:00
```

dgraham: This custom element is essentially a declarative way of exposing `Intl.DateTimeFormat` to markup, so it might be good to match the attribute names to the key names used by the formatter. In this case, it would be `timeZoneName`, even though that's not quite as nice as `timezone`.

mislav: Agreed!

```html
timeZoneName="short"
timeZoneName="long"
```

PR #73, opened 2016-11-22
Custom elements v1

tuespetre:
This is just some groundwork for migrating to the Custom Elements v1 spec. Do with it what you will

### DIFF
--- a/time-elements.js
+++ b/time-elements.js
@@ -1,9 +1,10 @@
-(function() {
+(function () {
   'use strict';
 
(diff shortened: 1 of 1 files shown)


tuespetre:
I now see #67 was opened last night (my time). Nevermind 😸

josh:
Ah, sorry about that. 

Yeah, this is going to be more complicated that I originally thought. You can't do `HTMLElement.call(this)` in any browser.

Thanks though!

tuespetre:
It's got to be shimmed by a polyfill unless you can go with pure ES2015 classes unfortunately.

I haven't tried it with document-register-element but it looks like it should support it too.

PR #74, opened 2016-11-22
Use phantomjs-prebuilt

josh:
Switches custom phantomjs fetching code to caching the npm phantomjs-prebuilt package.

* Sets up phantomjs binary locally on macOS when missing. For an example, I don't have phantomjs installed via homebrew.
* Ensures consistent phantomjs version locally and in CI.
* Less code.

Speed seems uneffected.

**master**

`npm install` (w/o phantomjs-prebuilt package) 12.44s

**phantomjs-cache**

Uncached `npm install` 21.85s
Cached `npm install`13.23s

##

CC: @mislav

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,5 @@
 language: node_js
 node_js: 6.9.1 # latest LTS
+cache:
+  directories:
+    - node_modules/phantomjs-prebuilt
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,2 @@
 language: node_js
 node_js: 6.9.1 # latest LTS
-cache:
-  directories:
-    - phantomjs
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ REPORTER = $(if $(CI),spec,dot)
 build: node_modules/
 
 test: node_modules/ build lint
-	phantomjs ./node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js \
+	./node_modules/phantomjs-prebuilt/bin/phantomjs ./node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js \
 		./test/test.html $(REPORTER) '{"useColors":true}'
 
 lint: node_modules/
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,6 @@
 REPORTER = $(if $(CI),spec,dot)
 
-ifeq ($(TRAVIS),)
 build: node_modules/
-else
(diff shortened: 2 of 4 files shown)


mislav:
I was considering phantomjs-prebuilt, but this is what [they say in the readme](https://github.com/Medium/phantomjs#continuous-integration):

> Please do not download PhantomJS for every CI job because it can quickly overload our CDNs. Instead take advantage of CI caching.

Then they go on to suggest that one should set up Travis caching exacly like I did. So in the end, if we listen to their advice and have a separate caching strategy, CI jobs have really no benefit of phantomjs-prebuilt.

.travis.yml
```diff
@@ -2,4 +2,4 @@ language: node_js
 node_js: 6.9.1 # latest LTS
 cache:
   directories:
-    - phantomjs
+    - node_modules/phantomjs-prebuilt
```
mislav: When cache is restored, this will create a `node_modules` directory. Then, `make build` is a no-op and will never run `npm install`.

josh: Travis invokes `npm install` directly by default for any node projects. `npm install` will only skip install the `phantomjs-prebuilt` package.

https://travis-ci.org/github/time-elements/builds/178111851#L149

josh:
> Please do not download PhantomJS for every CI job because it can quickly overload our CDNs. Instead take advantage of CI caching.

Yeah, this PR maintains binary caching.

Makefile
```diff
@@ -1,14 +1,9 @@
...
-ifeq ($(TRAVIS),)
 build: node_modules/
-else
-PATH := $(PWD)/phantomjs/bin:$(PATH)
-build: node_modules/ phantomjs/bin/phantomjs
-endif
 
 test: node_modules/ build lint
-	phantomjs ./node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js \
+	./node_modules/phantomjs-prebuilt/bin/phantomjs ./node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js \
```
mislav: I was going to suggest that this be `./node_modules/.bin/phantomjs`, but if the `phantomjs-prebuilt` directory is restored from cache, I'm not sure if `.bin/phantomjs` will even exist.

josh: Yeah, I actually hit that bug. fcd056b5b9a5befb7483e16bdd1ff2f997ec93cd made the change. Appears that `npm install` only rebuilds binstubs on packages it installs and does not rebuild them all. Though, I'm sure we could maybe pass some extra flag to it, but this seems fine.

PR #76, opened 2016-11-24
Add support `timeZoneName`

mislav:
Adding the `timeZoneName="long"` or `timeZoneName="short"` attribute adds the current timezone representation to the text visible to the user.

This is useful for situations in which the UI needs to make explicit to the user that the time is rendered in their own local timezone, rather than an arbitrary one determined by the server.

This is a simpler take at #65

/cc @stoe


test/local-time.js
```diff
@@ -66,4 +66,12 @@ suite('local-time', function() {
     }
     assert.include(['1969', '1970'], root.children[0].textContent)
   })
+
+  ;('Intl' in window ? test : test.skip)('displays time zone name', function() {
+    var time = document.createElement('local-time')
+    time.setAttribute('datetime', '1970-01-01T00:00:00.000Z')
+    time.setAttribute('minute', '2-digit')
+    time.setAttribute('timeZoneName', 'short')
```
josh: HTML attributes should be hyphen delimited.

mislav: @dgraham made a point about keeping the attribute names matching the properties in DateTimeFormat: https://github.com/github/time-elements/pull/65#r73983580

I could go either way.

dgraham: I suppose it is strange to have a mixed case attribute name when all others are lowercase. I'm fine with `<local-time time-zone-name="">` or `<local-time timezonename="">`. The property would still be camel cased as `timeZoneName` to match the formatter.

mislav: I would be most in favor for `timezone="short"` attribute, since `timezonename` or `time-zone-name` could be a mouthful. What do you think?

josh:
Maybe out of scope for this PR, but we should support property aliases on all this time formats so el.timeZoneName = writes to the attribute.

mislav:
> we should support property aliases on all this time formats so el.timeZoneName = writes to the attribute.

Declaring those properties on the element (if I understood you correctly) is a good idea. However, if the attribute name for this is `timezone` (that maps to `timeZoneName` parameter of DateTimeFormat), should the property name on the element be `timezone` or `timeZoneName`?

josh:
I like using timeZoneName as the property and time-zone-name as the attribute. That maintains a consistent mapping between camelcase and dasherized names.

josh:
For reference, `dataset`'s `DOMStringMap` https://html.spec.whatwg.org/multipage/dom.html#domstringmap

PR #78, opened 2017-03-08
Add internationalization using i18next

jcoyne:


jcoyne:
I think I figured out my build/test errors.  Please review this for me.

josh:
Thanks for the PR @jcoyne! But the direction we're planning to take here is to lean on the new `Intl.RelativeTimeFormat` for internationalization https://github.com/github/time-elements/pull/104. Thanks!

PR #86, opened 2017-11-11
mention another alternative lib

SimonWaldherr:
mention another alternative with support for many languages


dgraham:
This section of the readme is mostly intended to give credit to projects that predated this one, rather than providing a full listing of alternatives. Using npm package keywords is probably better for finding related projects.

PR #87, opened 2017-12-14
Chrome Headless, eslint upgrade, npm scripts

muan:
1. Remove phantomjs and add Karma
2. Remove makefile and use npm scripts
3. Upgrade eslint-plugin-github and prettify

💚 https://travis-ci.org/github/time-elements/builds/316327437

cc @github/web-systems


test/karma.config.js
```diff
@@ -0,0 +1,23 @@
...
+    ],
+    reporters: ['progress'],
+    port: 9876,
+    client: {mocha: {ui: 'tdd'}},
+    colors: true,
+    logLevel: config.LOG_INFO,
+    browsers: ['ChromeHeadless'],
+    autoWatch: false,
+    singleRun: true,
+    concurrency: Infinity
```
mislav: …and beyond! ✊

test/karma.config.js
```diff
@@ -0,0 +1,23 @@
+module.exports = function(config) {
+  config.set({
+    frameworks: ['mocha', 'chai'],
+    files: [
+      '../dist/time-elements.js',
```
mislav: We used to load the "legacy" file for tests, now we load the main one. I suspect this was because PhantomJS was unable to handle the non-legacy build.

However, if we load this test suite in an older browser, such as IE 11, would it work with this configuration?

.travis.yml
```diff
@@ -1,5 +1,2 @@
 language: node_js
 node_js: 6.9.1 # latest LTS
```
dgraham: Setting this to "node" will give us the latest version.

```
node_js:
  - "node"
```

.travis.yml
```diff
@@ -1,5 +1,2 @@
...
-cache:
-  directories:
-    - node_modules/phantomjs-prebuilt
```
dgraham: We can cache the `node_modules/` directory.

https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Caching-with-npm

test/karma.config.js
muan: Good point. I don't think it would. Since the legacy file is the main file on the module, I have changed this here to test with the legacy build instead.

PR #89, opened 2017-12-20
v1 fixes

muan:
For #67. @josh I made 2 fixes here:

1. Update callback names
2. Define custom elements after use, in `polymer.html`. More on this 👇🏻

This is a super weird Chrome bug that took away hours of my life 😭. While testing I noticed that it looks broken:

It doesn't happen in Safari. Turns out a bug has been filed: https://bugs.chromium.org/p/chromium/issues/detail?id=739417 Here's another test case I made: http://jsbin.com/xizojipoji/1/edit?html,output It seems like the original `textContent` gets appended late when element is defined **before** use, so it can't be cleared/replaced because it didn't exist.

cc @dgraham cause we talked about this.

---

Anyways all looks good now I think! :sparkles: 

Do you know if there's an efficient way to test the `setInterval` update for `relative-time`? The broken callbacks didn't get caught by the test.

### DIFF
--- a/examples/polymer.html
+++ b/examples/polymer.html
@@ -3,7 +3,6 @@
 <head>
   <meta charset="utf-8">
   <script src="https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.23/webcomponents.min.js"></script>
-  <script src="../dist/time-elements-legacy.js"></script>
 </head>
 <body>
   <h2>Past Date</h2>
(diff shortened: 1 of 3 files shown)


muan:
Oops forgot about this: `title` is being set repeatedly because `setAttribute('title', title)` triggers the callback.



e80897f fixes it.

tuespetre:
Oh man that Chrome bug has been :bug: bugging me!

PR #92, opened 2018-03-09
🔍check if `<local-time />` is defined before registering

koddsson:
This patch checks if a `<local-time />` element already exists before we register it on the page and only registers the element if there isn't one already registered.


koddsson:
>This works for me. Should we make this a standard pattern for all our other projects that publish custom elements?

I think so. We are already doing this in not published components so I think this is what we want.

PR #93, opened 2018-03-21
`datetime` changes update `title`

### LINKED ISSUE #56: Title attribute doesn't update on changes made to the datetime attribute
Check this Plunkr: http://plnkr.co/edit/xEB6WxUpfxgv41Aai4np?p=preview

When I update the date the text changes but the title doesn't.

I saw that it is because the title attribute take precedence, but I'm not sure it's a good case when the user didn't add a title.

Westbrook:
Fixes #56

### DIFF
--- a/src/extended-time-element.js
+++ b/src/extended-time-element.js
@@ -15,7 +15,7 @@ export default class ExtendedTimeElement extends HTMLElement {
     }
 
     const title = this.getFormattedTitle()
-    if (title && !this.hasAttribute('title')) {
+    if (title && (attrName === 'datetime' || !this.hasAttribute('title'))) {
       this.setAttribute('title', title)
     }
 


Westbrook:
When I look at the test that fails now, I am struck with reasoning for why you might want to keep the set `title`. However, this approach makes trouble when using a virtualized view like [`iron-list`](https://github.com/polymerelements/iron-list) where you recycle elements that have already had their title set internally which means they are displaying in correct dates.

Some quality of `_oldFormattedDate ` needs to be tracked. Would it be crazy to do just that where `this._date` is set?

Westbrook:
Possibly like:
```
    if (attrName === 'datetime') {
      const millis = Date.parse(newValue);
      if (this._date) {
        const oldTitle = this.getFormattedTitle();
      }
      this._date = isNaN(millis) ? null : new Date(millis);
    }

    const title = this.getFormattedTitle();
    const hasTitle = this.hasAttribute('title');
    const attrTitle = this.getAttribute('title');
    if (title && (!hasTitle || attrTitle === oldTitle) {
      this.setAttribute('title', title);
    }
```

muan:
@Westbrook Sorry for the unresponsiveness. I've merged #129 which addressed this issue using your suggested approach. It's released in v3.0.9. Thanks!

PR #96, opened 2018-07-25
Cache DateTimeFormat objects

dgraham:
Calling this constructor can account for 50% of script load time on pages with many custom time elements.

screen shot 2018-07-25 at 09 03 07

This is most noticeable on issue pages with many comments or large repository file listing pages where there's a brief flash of the raw date before the custom elements are upgraded.


src/utils.js
```diff
@@ -91,18 +91,24 @@ export function strftime(time, formatString) {
...
 export function makeFormatter(options) {
-  if ('Intl' in window) {
-    try {
-      return new Intl.DateTimeFormat(undefined, options)
-    } catch (e) {
-      if (!(e instanceof RangeError)) {
-        throw e
+  let format
+  return function() {
+    if (format) return format
```
mislav: What if different options are passed than ones this was initialized with?

dgraham: A simplified calling example:

```js
const fn = makeFormatter({options here})
const formatter = fn() // no options here
formatter.format(new Date())
```

The returned factory function is a closure over the `let format` binding, so it's unique to each returned function. The `options` aren't overridden by subsequent calls.

Another approach would be a map of options to formatters.

```js
const formatters = {}
function makeFormatter(options) {
  const cached = formatters[options]
  if (cached) return cached

  // This constructor can throw.
  const format = new Intl.DateTimeFormat(options)
  formatters[options] = format
  return format
}
```

The closure-as-cache preserves the existing error handling though, preventing an exception in top-level module code from crashing the bundle on page load.

Issue #98, opened 2018-10-23
Implement RelativeTime on Intl.RelativeTimeFormat API


josh:
https://developers.google.com/web/updates/2018/10/intl-relativetimeformat
https://github.com/tc39/proposal-intl-relative-time

##

CC: @github/web-systems

muan:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Browser_compatibility

PR #103, opened 2019-04-11
Add tests for special cases in relative-time

koddsson:
Adding these tests since the `Intl.RelativeTimeFormat` API doesn't support messages such as `a minute ago`. Having these tests will be useful so that we're aware of those breaking changes when we implement #98.

Ref: https://github.com/tc39/proposal-intl-relative-time#example-1

### DIFF
--- a/test/relative-time.js
+++ b/test/relative-time.js
@@ -1,4 +1,46 @@
 suite('relative-time', function() {
+  test('rewrites from now past datetime to days ago', function() {
+    const now = new Date(Date.now() - 3 * 60 * 60 * 24 * 1000).toISOString()
+    const time = document.createElement('relative-time')
+    time.setAttribute('datetime', now)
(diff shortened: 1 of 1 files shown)


test/relative-time.js
```diff
@@ -1,4 +1,46 @@
...
+  })
+
+  test('rewrites from now future datetime to minutes from now', function() {
+    const now = new Date(Date.now() + 3 * 60 * 60 * 1000).toISOString()
+    const time = document.createElement('relative-time')
+    time.setAttribute('datetime', now)
+    assert.equal(time.textContent, '3 hours from now')
+  })
+
+  test('rewrites from now past datetime to a hour ago', function() {
```
dgraham:
```suggestion
  test('rewrites from now past datetime to an hour ago', function() {
```

PR #104, opened 2019-04-12
Use Intl.RelativeTimeFormat

josh:
Aims to implement https://github.com/github/time-elements/issues/98.

**To-do**

* [x] Feature detect `Intl.RelativeTimeFormat`.
* [ ] Decide how to handle "micro" formatter.
* [x] Fix tests.
* [ ] Implement `locale` attribute

### DIFF
--- a/src/relative-time.js
+++ b/src/relative-time.js
@@ -1,6 +1,6 @@
 /* @flow strict */
 
-import {strftime, makeFormatter, isDayFirst, isThisYear, isYearSeparator} from './utils'
+import {strftime, makeFormatter, makeRelativeFormatter, isDayFirst, isThisYear, isYearSeparator} from './utils'
 
 export default class RelativeTime {
   date: Date
@@ -62,31 +62,29 @@ export default class RelativeTime {
     const month = Math.round(day / 30)
     const year = Math.round(month / 12)
     if (ms < 0) {
-      return 'just now'
+      return formatRelativeTime(0, 'second')
     } else if (sec < 10) {
-      return 'just now'
+      return formatRelativeTime(0, 'second')
     } else if (sec < 45) {
-      return `${sec} seconds ago`
+      return formatRelativeTime(-sec, 'second')
     } else if (sec < 90) {
-      return 'a minute ago'
+      return formatRelativeTime(-min, 'minute')
     } else if (min < 45) {
-      return `${min} minutes ago`
+      return formatRelativeTime(-min, 'minute')
     } else if (min < 90) {
-      return 'an hour ago'
+      return formatRelativeTime(-hr, 'hour')
     } else if (hr < 24) {
-      return `${hr} hours ago`
+      return formatRelativeTime(-hr, 'hour')
     } else if (hr < 36) {
-      return 'a day ago'
+      return formatRelativeTime(-day, 'day')
     } else if (day < 30) {
-      return `${day} days ago`
+      return formatRelativeTime(-day, 'day')
     } else if (day < 45) {
-      return 'a month ago'
-    } else if (month < 12) {
-      return `${month} months ago`
+      return formatRelativeTime(-month, 'month')
     } else if (month < 18) {
-      return 'a year ago'
+      return formatRelativeTime(-year, 'year')
     } else {
-      return `${year} years ago`
+      return formatRelativeTime(-year, 'year')
     }
   }
 
@@ -124,29 +122,29 @@ export default class RelativeTime {
     const month = Math.round(day / 30)
     const year = Math.round(month / 12)
     if (month >= 18) {
-      return `${year} years from now`
+      return formatRelativeTime(year, 'year')
     } else if (month >= 12) {
-      return 'a year from now'
+      return formatRelativeTime(year, 'year')
     } else if (day >= 45) {
-      return `${month} months from now`
+      return formatRelativeTime(month, 'month')
     } else if (day >= 30) {
-      return 'a month from now'
+      return formatRelativeTime(month, 'month')
     } else if (hr >= 36) {
-      return `${day} days from now`
+      return formatRelativeTime(day, 'day')
     } else if (hr >= 24) {
-      return 'a day from now'
+      return formatRelativeTime(day, 'day')
     } else if (min >= 90) {
-      return `${hr} hours from now`
+      return formatRelativeTime(hr, 'hour')
     } else if (min >= 45) {
-      return 'an hour from now'
+      return formatRelativeTime(hr, 'hour')
     } else if (sec >= 90) {
-      return `${min} minutes from now`
+      return formatRelativeTime(min, 'minute')
     } else if (sec >= 45) {
-      return 'a minute from now'
+      return formatRelativeTime(min, 'minute')
     } else if (sec >= 10) {
-      return `${sec} seconds from now`
+      return formatRelativeTime(sec, 'second')
     } else {
-      return 'just now'
+      return formatRelativeTime(0, 'second')
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/relative-time.js
```diff
@@ -62,31 +64,29 @@ export default class RelativeTime {
     const month = Math.round(day / 30)
     const year = Math.round(month / 12)
     if (ms < 0) {
-      return 'just now'
+      return rtf.format(0, 'second')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(0, 'second')
//=> "now"
```

src/relative-time.js
```diff
@@ -62,31 +64,29 @@ export default class RelativeTime {
...
     } else if (sec < 10) {
-      return 'just now'
+      return rtf.format(0, 'second')
     } else if (sec < 45) {
-      return `${sec} seconds ago`
+      return rtf.format(-sec, 'second')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(-30, 'second')
//=> "30 seconds ago"
```

src/relative-time.js
```diff
@@ -62,31 +64,29 @@ export default class RelativeTime {
...
     } else if (sec < 90) {
-      return 'a minute ago'
+      return rtf.format(-min, 'minute')
     } else if (min < 45) {
-      return `${min} minutes ago`
+      return rtf.format(-min, 'minute')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(-30, 'minute')
//=> "30 minutes ago"
```

src/relative-time.js
```diff
@@ -62,31 +64,29 @@ export default class RelativeTime {
...
     } else if (min < 90) {
-      return 'an hour ago'
+      return rtf.format(-hr, 'hour')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(-30, 'hour')
//=> "30 hours ago"
```

src/relative-time.js
```diff
@@ -62,31 +64,29 @@ export default class RelativeTime {
...
     } else if (hr < 24) {
-      return `${hr} hours ago`
+      return rtf.format(-hr, 'hour')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(-30, 'hour')
//=> "30 hours ago"
```

src/relative-time.js
```diff
@@ -62,31 +64,29 @@ export default class RelativeTime {
...
+      return rtf.format(0, 'second')
     } else if (sec < 10) {
-      return 'just now'
+      return rtf.format(0, 'second')
     } else if (sec < 45) {
-      return `${sec} seconds ago`
+      return rtf.format(-sec, 'second')
     } else if (sec < 90) {
-      return 'a minute ago'
+      return rtf.format(-min, 'minute')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(-1, 'minute')
//=> "1 minute ago"
```

Loss of "an".

src/relative-time.js
```diff
@@ -2,6 +2,8 @@
 
 import {strftime, makeFormatter, isDayFirst, isThisYear, isYearSeparator} from './utils'
 
+const rtf = new Intl.RelativeTimeFormat('en', {numeric: 'auto'})
```
josh: Needs feature detection and fallback.

src/relative-time.js
```diff
@@ -124,29 +124,29 @@ export default class RelativeTime {
     const month = Math.round(day / 30)
     const year = Math.round(month / 12)
     if (month >= 18) {
-      return `${year} years from now`
+      return rtf.format(year, 'year')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(2, 'year')
//=> "in 2 years"
```

src/relative-time.js
```diff
@@ -124,29 +124,29 @@ export default class RelativeTime {
...
+      return rtf.format(month, 'month')
     } else if (day >= 30) {
-      return 'a month from now'
+      return rtf.format(month, 'month')
     } else if (hr >= 36) {
-      return `${day} days from now`
+      return rtf.format(day, 'day')
     } else if (hr >= 24) {
-      return 'a day from now'
+      return rtf.format(day, 'day')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(1, 'day')
//=> "tomorrow"
```

src/relative-time.js
```diff
@@ -124,29 +124,29 @@ export default class RelativeTime {
...
+      return rtf.format(year, 'year')
     } else if (month >= 12) {
-      return 'a year from now'
+      return rtf.format(year, 'year')
     } else if (day >= 45) {
-      return `${month} months from now`
+      return rtf.format(month, 'month')
     } else if (day >= 30) {
-      return 'a month from now'
+      return rtf.format(month, 'month')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(1, 'month')
//=> "next month"
```

src/relative-time.js
```diff
@@ -124,29 +124,29 @@ export default class RelativeTime {
...
+      return rtf.format(hr, 'day')
     } else if (min >= 45) {
-      return 'an hour from now'
+      return rtf.format(hr, 'day')
     } else if (sec >= 90) {
-      return `${min} minutes from now`
+      return rtf.format(min, 'minute')
     } else if (sec >= 45) {
-      return 'a minute from now'
+      return rtf.format(min, 'minute')
```
josh:
```js
new Intl.RelativeTimeFormat('en', {numeric: 'auto'}).format(1, 'minute')
//=> "in 1 minute"
```

src/relative-time.js
```diff
@@ -2,6 +2,8 @@
 
 import {strftime, makeFormatter, isDayFirst, isThisYear, isYearSeparator} from './utils'
 
+const rtf = new Intl.RelativeTimeFormat('en', {numeric: 'auto'})
+
 export default class RelativeTime {
```
josh: This `RelativeTime` class helper probably needs to say around to wrap are rules about which unit precision we prefer to display.

josh:
doh, @koddsson did you already start working on this? (https://github.com/github/time-elements/pull/103)

koddsson:
@josh Nah I just started exploring and noticed that some tests for special cases that would break when we'd start using `RelativeTimeFormat` so I added those 😸

src/relative-time.js
koddsson: Nice! I looked at the list [here](https://github.com/tc39/proposal-intl-relative-time#example-1) and didn't find a substitute to "Just now" so I wasn't sure how we were going to solve that 😄 

https://github.com/tc39/proposal-intl-relative-time/pull/107

mislav:
1. Is there going to be a `locale` attribute so that users may pass a locale specifier other than "en" down to Intl?
2. If there isn't a `locale` attribute, should we attempt to read the default locale from the document, e.g. from `<html lang="...">`? https://www.w3.org/International/questions/qa-html-language-declarations

keithamus:
> should we attempt to read the default locale from the document 

`navigator.language` can be used

josh:
Alright, got tests passing here. Could use some review please 😁

test/relative-time.js
```diff
@@ -31,14 +31,14 @@ suite('relative-time', function() {
     const now = new Date(Date.now() + 3 * 60 * 60 * 1000).toISOString()
     const time = document.createElement('relative-time')
     time.setAttribute('datetime', now)
-    assert.equal(time.textContent, '3 hours from now')
+    assert.equal(time.textContent, 'in 3 days')
```
dgraham: Is the change in units from hours to days expected?

josh: good catch. I don't think so. let me take a look why this happened.

josh:
Would also love some feedback on the effect of this change for people who can speak more than English 😁

src/utils.js
```diff
@@ -164,3 +164,23 @@ export function isThisYear(date: Date) {
...
+// eslint-disable-next-line flowtype/no-weak-types
+export function makeRelativeFormatter(options: any): () => ?any {
+  let format
+  return function() {
+    if (format) return format
+    if ('Intl' in window && 'RelativeTimeFormat' in window.Intl) {
+      try {
+        // eslint-disable-next-line flowtype/no-flow-fix-me-comments
+        // $FlowFixMe: missing RelativeTimeFormat type
+        format = new Intl.RelativeTimeFormat(undefined, options)
```
dgraham: I think we want to pass `en` as the locale until we add a `locale` attribute to the custom element so it's configurable by the host app. That will prevent this text from appearing in the default locale for the browser with the rest of the text appearing in English.

test/relative-time.js
josh: Yep, bug

https://github.com/github/time-elements/blob/2f3e55029379817c7887b34755712f5b253f130d/src/relative-time.js#L136-L139

src/utils.js
josh: Yeah, I'm down to support that, though I was just stealing from our existing `makeFormatter`. 

https://github.com/github/time-elements/blob/4e6f95a34fc897626e794ed78f0f45701043de67/src/utils.js#L102

Will that need to change as well? Should I do this in a separate PR? Starting to wonder how this change is going to ship. It might be a major version bump. So could bundle this and that locale change in one release.

josh:
> Decide how to handle "micro" formatter.

Not sure what to do, or even if we have to deal with this.

src/utils.js
dgraham: I wonder if the [runtime's default locale](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation) is hardcoded to `en` in browsers and this is different from the user's preferred language setting in the operating system. Anyway, since undefined is already working for DateTimeFormat, we can use it here too.

A major version release would be appropriate since this changes the text output slightly.

We could couple that with `<relative-time locale="…">` if it's not too much work. The only weird thing is that the locale attribute would only apply to browsers with `RelativeTimeFormat` support. The older code path would still use the hardcoded English text.

josh:
Thanks! I'll move forward with merging this and open another PR attempt for the `locale=` attribute.

PR #105, opened 2019-04-17
rm some whitespace

steventam312:


koddsson:
Hey @steventam312. These indents are intentional for readability.

PR #106, opened 2019-05-14
Add flow definition file

koddsson:
https://github.com/github/web-systems/issues/215

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -1,0 +1,25 @@
+/* flow strict */
+
+declare module 'time-elements' {
+  declare class ExtendedTimeElement extends HTMLElement {
+    _date: ?Date;
+    getFormattedTitle(): ?string;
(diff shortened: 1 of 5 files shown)


src/index.js.flow
```diff
@@ -0,0 +1,25 @@
+/* flow strict */
+
+declare module 'time-elements' {
+  declare class ExtendedTimeElement extends HTMLElement {
+    _date: ?Date;
```
mislav: Should we avoid exposing something that is practically a private property?

koddsson: I'll remove it from the definition for now but I think it begs the question if we want to just leave properties off the flow definition if it shouldn't be part of the public API. The property will still be available at run-time but for applications using flow it will cause linting errors.

koddsson:
```suggestion
```

PR #107, opened 2019-05-16
Expose faux private field 

koddsson:
At github we are trying to extend one of the elements and using it in application code but flow keeps complaining that `_date` isn't defined on the element since we decided not to ship it with the flow definitions

This PR adds the property back in so we can access it and unblock the PR.

Ref: https://github.com/github/time-elements/pull/106#discussion_r283850992
Ref: https://github.com/github/github/pull/115283#issuecomment-492978635

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -2,6 +2,7 @@
 
 declare module 'time-elements' {
   declare class ExtendedTimeElement extends HTMLElement {
+    _date: ?Date;
     getFormattedTitle(): ?string;
     getFormattedDate(): ?string;
   }


keithamus:
I'm curious as to why `_date` is faux private in the first place. What benefit does it have to be faux private? Is it a case that we don't want it to become out of sync with the representation? In that case why not make it a getter/setter?

koddsson:
> I'm curious as to why `_date` is faux private in the first place. What benefit does it have to be faux private? Is it a case that we don't want it to become out of sync with the representation? In that case why not make it a getter/setter?

I think that's a good question and the answer might be that it's due for a refactor. Changing the property to be getter/setter that will update the attributes on the element would probably be the way to go.

In the meantime I'm gonna merge and publish this change to unblock people 💃

dgraham:
Can we disable Flow checks in the few places `_date` is referenced instead?

koddsson:
> Can we disable Flow checks in the few places `_date` is referenced instead?

Yeah I think that could also work but you then have the drawback of `_date` being typed as `any`.

PR #108, opened 2019-05-17
Change faux private property to a getter

koddsson:
This will mean that application code cannot set the date directly since there is no setter defined. 

Ref: #107

### DIFF
--- a/src/extended-time-element.js
+++ b/src/extended-time-element.js
@@ -3,21 +3,13 @@
 import {makeFormatter} from './utils'
 
 export default class ExtendedTimeElement extends HTMLElement {
-  _date: ?Date
-
   static get observedAttributes() {
     return ['datetime', 'day', 'format', 'hour', 'minute', 'month', 'second', 'title', 'weekday', 'year']
   }
 
   // Internal: Refresh the time element's formatted date when an attribute changes.
-  //
-  // Returns nothing.
+  // eslint-disable-next-line no-unused-vars
   attributeChangedCallback(attrName: string, oldValue: string, newValue: string) {
-    if (attrName === 'datetime') {
-      const millis = Date.parse(newValue)
-      this._date = isNaN(millis) ? null : new Date(millis)
-    }
-
     const title = this.getFormattedTitle()
     if (title && !this.hasAttribute('title')) {
       this.setAttribute('title', title)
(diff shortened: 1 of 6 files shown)


src/extended-time-element.js
```diff
@@ -3,21 +3,13 @@
...
 
   // Internal: Refresh the time element's formatted date when an attribute changes.
-  //
-  // Returns nothing.
+  // eslint-disable-next-line no-unused-vars
   attributeChangedCallback(attrName: string, oldValue: string, newValue: string) {
-    if (attrName === 'datetime') {
-      const millis = Date.parse(newValue)
-      this._date = isNaN(millis) ? null : new Date(millis)
-    }
```
dgraham: The nice thing about the callback is that we're parsing the date only once and caching its value until it's changed. Depending on how frequently the new getter method is called the date parsing could show up in performance timing graphs, similar to #96.

keithamus: I was curious to see if Date.parse() was really that slow compared to getting from a WeakMap; I have to say [the results are surprising](https://jsperf.com/parsing-date-vs-caching-in-weakmap). WeakMap is not much faster for unhappy paths (I presume because `Date.parse()` knows how to bail out quickly). Of course I'd say given how trivial it is for us to implement a WeakMap here, we should.

src/time-ago-element.js
```diff
@@ -6,11 +6,11 @@ import RelativeTimeElement from './relative-time-element'
 export default class TimeAgoElement extends RelativeTimeElement {
   getFormattedDate(): ?string {
     const format = this.getAttribute('format')
-    if (!this._date) return
+    if (!this.date) return
```
keithamus: We could cache `this.date` here

src/time-until-element.js
```diff
@@ -6,11 +6,11 @@ import RelativeTimeElement from './relative-time-element'
 export default class TimeUntilElement extends RelativeTimeElement {
   getFormattedDate(): ?string {
     const format = this.getAttribute('format')
-    if (!this._date) return
+    if (!this.date) return
```
keithamus: We could cache this.date here

src/relative-time-element.js
```diff
@@ -5,8 +5,8 @@ import ExtendedTimeElement from './extended-time-element'
 
 export default class RelativeTimeElement extends ExtendedTimeElement {
   getFormattedDate(): ?string {
-    if (this._date) {
-      return new RelativeTime(this._date).toString()
+    if (this.date) {
```
keithamus: We could cache this.date here

josh:
Release note: I just merged some backwards incompatible changes into `master`. So `master` will be targeting a `3.x` release. This still makes sense to merge there, but maybe we'll want to consider backporting this change to a `2.x` as well.

koddsson:
> maybe we'll want to consider backporting this change to a `2.x` as well.

Since this is a breaking change for apps that are utilizing the `_date` property, would it make sense to backport to `2.x`?

PR #109, opened 2019-05-20
Support lang attribute

josh:
The intent here is to respect any `HTMLElement.lang` attribute set on a time element or it's parents when rendering localized text.

Issues/questions:

* [x] Does this only affect relative time related elements? Can it be used with local time elements at all?
* [x] Travis CI falls as it's environment doesn't support any locales and falls back to english. Though, this works locally for me. Not sure how to best write the tests in a feature detected way.
* [x] My plumbing of the locale string isn't super elegant.


test/relative-time.js
```diff
@@ -159,4 +159,22 @@ suite('relative-time', function() {
...
+    assert.equal(time.getFormattedDate(), 'hace 3 días')
+  })
+
+  test('rewrites given parent lang attribute', function() {
+    const container = document.createElement('span')
+    container.setAttribute('lang', 'es')
+    const now = new Date(Date.now() - 3 * 60 * 60 * 24 * 1000).toISOString()
+    const time = document.createElement('relative-time')
+    time.setAttribute('datetime', now)
+    container.appendChild(time)
```
mislav: At this time, the contents of the `relative-time` element is already rendered, and re-parenting it (so it inherits the parent's `lang` attribute) won't change the already rendered contents, I think. Perhaps for the sake of the test, you could do `setAttribute('datetime', ...)` after it has been added to the correct parent?

test/relative-time.js
```diff
@@ -159,4 +159,22 @@ suite('relative-time', function() {
     }
     assert.equal(root.children[0].textContent, 'now')
   })
+
+  test('rewrites given lang attribute', function() {
+    const now = new Date(Date.now() - 3 * 60 * 60 * 24 * 1000).toISOString()
+    const time = document.createElement('relative-time')
+    time.setAttribute('datetime', now)
+    time.setAttribute('lang', 'es')
```
mislav: For this to work, maybe you would have to add `lang` to `observedAttributes`

src/relative-time.js
```diff
@@ -187,8 +189,8 @@ export default class RelativeTime {
   }
 }
 
-function formatRelativeTime(value: number, unit: string): string {
-  const formatter = relativeFormatter()
+function formatRelativeTime(locale: string, value: number, unit: string): string {
+  const formatter = makeRelativeFormatter(locale, {numeric: 'auto'})()
```
dgraham: The formatter instance caching in `makeRelativeFormatter` isn't really effective since we invoke the factory function each time here. Maybe we should separate the factory from the caching and just call the factory here so it's clear it isn't attempting to cache it.

```js
// eslint-disable-next-line flowtype/no-weak-types
export function makeRelativeFormatter(options: any): () => ?any {
  let format
  return function() {
    if (format) return format
    format = makeRelativeFormat(options)
  }
}

export function makeRelativeFormat(options: any): ?Intl.RelativeTimeFormat {
  if ('Intl' in window && 'RelativeTimeFormat' in window.Intl) {
    try {
      // eslint-disable-next-line flowtype/no-flow-fix-me-comments
      // $FlowFixMe: missing RelativeTimeFormat type
      return new Intl.RelativeTimeFormat(undefined, options)
    } catch (e) {
      if (!(e instanceof RangeError)) {
        throw e
      }
    }
  }
}
```

test/relative-time.js
josh: Done. Though, one issue here is that we can't easily observe the attributes of all our parents. If the document body's lang attribute changed, our children won't notice.

src/utils.js
```diff
@@ -166,21 +166,27 @@ export function isThisYear(date: Date) {
...
-      try {
-        // eslint-disable-next-line flowtype/no-flow-fix-me-comments
-        // $FlowFixMe: missing RelativeTimeFormat type
-        format = new Intl.RelativeTimeFormat(undefined, options)
-        return format
-      } catch (e) {
-        if (!(e instanceof RangeError)) {
-          throw e
-        }
+export function makeRelativeFormat(locale: string, options: any): ?any {
```
mislav: I understand `any` wasn't added or changed in this PR, but could we get rid of it here? We could define a type that declares the specific parameters that we support and forward to `Intl.RelativeTimeFormat`, and return a type that basically just implements the `format()` method (since that's the only method that we interface with).

We can later drop all of that when Flow implements `Intl.RelativeTimeFormat` natively.

josh:
I think I integrated everyone's feedback. Thanks! 💖 

Could I get a final set of 👀s once more?

Issue #111, opened 2019-06-02
Mixed languages in '<relative-time>'


DominikPalo:
Issues lists on GitHub have relative times displayed in two languages. For example, my os/browser is set to the Slovak language and when I open a GitHub issues list, some relative times are displayed as translated (red) and some in English (blue) - see the attached screenshot.

Browser: Google Chrome 74.0.3729.169
OS: macOS Mojave 10.14.4 

<img width="485" alt="github" src="https://user-images.githubusercontent.com/1623257/58759628-5c4bba00-852d-11e9-8e4a-ac9b37f8e162.png">

Faierbel:
The same problem occurs with me. Part of it is in Polish, in part in English.
![Annotation 2019-06-02 134839](https://user-images.githubusercontent.com/25534725/58760829-68d80e80-853d-11e9-9ace-2ec20b3d9b06.png)

koddsson:
Hey @DominikPalo and @Faierbel 👋

I'm unable to reproduce this on my machine. Would you happen to have any Chrome extensions installed that could be causing this? Is this reproducible in Incognito Mode (or with extensions disabled).

koddsson:
Hey @DominikPalo and @Faierbel 👋

I'm unable to reproduce this on my machine. Would you happen to have any Chrome extensions installed that could be causing this? Is this reproducible in Incognito Mode (or with extensions disabled).

khellang:
I saw this 2 minutes ago as well (mix between Norwegian and English, didn't capture screenshot), but I'm unable to repro anymore 🤔 😕

DominikPalo:
I've just tried and it happens also in Incognito mode (with all Chrome extensions disabled). But I have to switch to the "Closed" tab and then back to the "Open" tab to replicate it - see the recording: https://gyazo.com/126ce5a0351f06af12f03d7116d69622

khellang:
Managed to reproduce after a few refreshes on pulls, but not on issues:

![image](https://user-images.githubusercontent.com/582487/58799096-a18eeb00-8604-11e9-96ae-2095575da26b.png)

khellang:
BTW, I have `lang="en"` on the `html` tag, `Accept-Language: en,nb;q=0.9` and `navigator.language === "en"` 🤷‍♂

koddsson:
I've figured out what is going on with all your help and am able to reproduce this. I'll try to have a fix going out soon!

koddsson:
[...]

PR #112, opened 2019-06-03
Refresh locale when element is connected to DOM

koddsson:
When time elements are added to the DOM via PJAX they are constructed without access to a parent element with a `lang` attribute so they are constructed with the locale `default`. 

This patch makes sure to re-set the text contents of the element when the element is connected to the DOM so that it can make sure to read any container elements that might have a `lang` attribute and display the correct locale.

Fixes: #111

### DIFF
--- a/src/extended-time-element.js
+++ b/src/extended-time-element.js
@@ -9,6 +9,18 @@ export default class ExtendedTimeElement extends HTMLElement {
     return ['datetime', 'day', 'format', 'lang', 'hour', 'minute', 'month', 'second', 'title', 'weekday', 'year']
   }
 
+  connectedCallback() {
+    const title = this.getFormattedTitle()
+    if (title && !this.hasAttribute('title')) {
+      this.setAttribute('title', title)
+    }
+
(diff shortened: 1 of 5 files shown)


koddsson:
The "failing test" I added is apparently never failing so I'm working on fixing that so that we have better confidence on shipping this.

josh:
> When time elements are added to the DOM via PJAX they are constructed without access to a parent element with a `lang` attribute so they are constructed with the locale `default`.

Ah! I was wondering why the `lang` attribute wasn't working for us.

koddsson:
I've spent a bunch of time trying to replicate the PJAX behaviour in the test suite but for some reason I'm not able 😢 

If someone in @github/web-systems wants to take a stab at it or pair with me on it that would be cool but for now I've dropped the test commit so only the fix is in this PR. When linking it to `github/github` it fixes the issue for me since `connectedCallback` is called when the element is connected to the document and it will now trigger a `closest` search for a element with a `lang` attribute and set that as the locale.

Issue #113, opened 2019-06-05
Relative time element shows "This year" instead of "x months ago"


koddsson:
In December "Last year" has a range of 1-12 months 😱 It would be great if we could make sure to show "x months ago" again.

Ref: https://github.com/github/github/issues/116284
Ref: https://github.com/github/github/pull/115903

PR #114, opened 2019-06-05
Make sure to format months ago correctly

koddsson:
I think we were erronusly only displaying "x months ago" when the time elapsed was 31-45 days. This patch makes sure we are showing "x months ago" when the time elapsed is 31days - 11 months.

Fixes: #113

### DIFF
--- a/test/time-ago.js
+++ b/test/time-ago.js
@@ -37,6 +37,20 @@ suite('time-ago', function() {
     assert.equal(root.children[0].textContent, 'now')
   })
 
+  test('rewrites from now past datetime to months ago', function() {
+    const now = new Date(Date.now() - 3 * 30 * 24 * 60 * 60 * 1000).toISOString()
+    const time = document.createElement('time-ago')
+    time.setAttribute('datetime', now)
+    assert.equal(time.textContent, '3 months ago')
+  })
+
+  test('rewrites from now past datetime to years ago', function() {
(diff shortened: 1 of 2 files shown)


test/time-ago.js
```diff
@@ -37,6 +37,20 @@ suite('time-ago', function() {
...
+    const time = document.createElement('time-ago')
+    time.setAttribute('datetime', now)
+    assert.equal(time.textContent, '3 months ago')
+  })
+
+  test('rewrites from now past datetime to years ago', function() {
+    const now = new Date(Date.now() - 12 * 30 * 24 * 60 * 60 * 1000).toISOString()
+    const time = document.createElement('time-ago')
+    time.setAttribute('datetime', now)
+    assert.equal(time.textContent, 'last year')
```
mislav: What does "last year" mean if it's January right now? Can we instead have it say something like “1 year ago”?

koddsson: It means "in the last 12 months" I believe.

The `last year` text comes from us setting the `numeric` option to `auto` and not the default `always`. It's the option that gives us back timestamps such as `yesterday`.

- [RelativeTimeFormat - Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RelativeTimeFormat#Parameters)

We can special case the time there to set the `numeric` option to `always`.

PR #116, opened 2019-06-13
Add missing @github scope to README and flow files

koddsson:
I feel a bit dumb missing this 😬

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -1,4 +1,4 @@
-/* flow strict */
+ /* @flow strict */
 
 declare module '@github/time-elements' {
   declare class ExtendedTimeElement extends HTMLElement {
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -1,4 +1,4 @@
- /* @flow strict */
+/* @flow strict */
 
 declare module '@github/time-elements' {
   declare class ExtendedTimeElement extends HTMLElement {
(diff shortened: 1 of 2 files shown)


src/index.js.flow
```diff
@@ -1,6 +1,6 @@
 /* flow strict */
```
koddsson:
```suggestion
/* @flow strict */
```

src/index.js.flow
```diff
@@ -1,6 +1,6 @@
-/* flow strict */
+ /* @flow strict */
```
mislav: Nit: space

src/index.js.flow
```diff
@@ -1,6 +1,6 @@
...
 
-declare module 'time-elements' {
+declare module '@github/time-elements' {
```
mislav: Can we drop `declare module` or at least make it anonymous so we don't have to keep updating it through renames?

koddsson: Yup. I'm linking it to `github/github` right now to make sure it works with the dropped `module` scope.

koddsson: We talked in slack but for posterity, we need this to be declared in a module since we include it in the `[libs]` section of our `.flowconfig` and we need to signal to `flow` what module this file is for.

src/index.js.flow
koddsson:
```suggestion
/* @flow strict */
```

PR #118, opened 2019-07-22
Add typescript definition file

koddsson:
@cheshire137 Can you test this in your project? I think this is correct but I'd love a second pair of 👀

Ref: https://github.com/github/web-systems/issues/223

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 declare class ExtendedTimeElement extends HTMLElement {
-  get date(): Date | undefined
+  readonly date: Date | undefined
   getFormattedTitle(): string | undefined
   getFormattedDate(): string | undefined
 }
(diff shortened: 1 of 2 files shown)


cheshire137:
What do you think is the best way to verify this? Save the files from this branch into the node_modules time-elements directory and see if the warnings go away when I import the package?

koddsson:
You can install the code from this PR directly like so:

```sh
npm install --save github/time-elements#add-typescript-definition-file
```

And after you are done testing you can go back to using the code from npm with:

```sh
npm remove @github/time-elements
npm install --save @github/time-elements
```

koddsson:
@cheshire137 Did you have a chance to check this out? If I can help in any way, just let me know ✨

koddsson:
Merging this! 👾

PR #129, opened 2019-11-07
Update title attribute when datetime attribute is changed

### LINKED ISSUE #56: Title attribute doesn't update on changes made to the datetime attribute

muan:
Supersedes #93, fixes #56.

There is an existing test for `title` not getting changed if there's already one set:

https://github.com/github/time-elements/blob/e763503ce57465f7f469f1a6642ea9dd581cb53e/test/title-format.js#L13-L18

This uses a WeakMap to track if `title` attribute was set by the element. If it was, future title attribute changes would be allowed.


muan:
I added a test case in https://github.com/github/time-elements/pull/129/commits/7e1ebdd1618e274e04be576dd254ae16a82d62ad, which prompted me to pursue another approach, and ended up just like what was proposed in #93. Please review again. Thanks!

PR #132, opened 2020-01-08
Export RelativeTime

dotpointer:
Export RelativeTime so details inside it can be used outside of it.


dotpointer:
Made in Neovici fork instead.

PR #145, opened 2021-03-18
refactor: simplify nowelements tracking with a set

keithamus:
The tracking for elements which need to be updated to change to `now` is a little more complex than it needs to be with modern JS. It's also the suspect in a memory leak we have for long running tabs.

My theory is a `connectedCallback` is fired twice as an element moves in the DOM, and the array gets populated. The array is `spliced` with an `indexOf` check, so if duplicates are in the array it will not clean them up, causing a memory leak.

The "solution" could be to do a `while(nowElements.includes(this))`... or just use a `Set` :laughing:.

This changes the code to use a `Set`, it also reconfigures the timeout code to ensure we avoid `setInterval` which can cause queueing issues, and only dispatch a `setTimeout` in the `updateNowElements` call. This simplifies the logic quite a bit.

### DIFF
--- a/src/relative-time-element.js
+++ b/src/relative-time-element.js
@@ -4,6 +4,21 @@ import RelativeTime from './relative-time'
 import ExtendedTimeElement from './extended-time-element'
 import {localeFromElement} from './utils'
 
+// Internal: Array tracking all elements attached to the document that need
+// to be updated every minute.
+const nowElements = new Set()
+
+// Internal: Timer ID for `updateNowElements` interval.
+let updateNowElementsId
+
+// Internal: Install a timer to refresh all attached relative-time elements every
+// minute.
+function updateNowElements() {
+  clearTimeout(updateNowElementsId)
+  for (const el of nowElements) el.textContent = el.getFormattedDate() || ''
+  if (nowElements.size) updateNowElementsId = setTimeout(updateNowElements, 60 * 1000)
+}
+
 export default class RelativeTimeElement extends ExtendedTimeElement {
   getFormattedDate(): ?string {
     const date = this.date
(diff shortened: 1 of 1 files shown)


src/relative-time-element.js
```diff
@@ -13,44 +28,13 @@ export default class RelativeTimeElement extends ExtendedTimeElement {
...
-let updateNowElementsId
-
-// Internal: Install a timer to refresh all attached relative-time elements every
-// minute.
-function updateNowElements() {
-  let time, i, len
-  for (i = 0, len = nowElements.length; i < len; i++) {
-    time = nowElements[i]
-    time.textContent = time.getFormattedDate() || ''
+    nowElements.remove(this)
```
koddsson, line 37:
```suggestion
    nowElements.delete(this)
```

src/relative-time-element.js
```diff
@@ -4,6 +4,21 @@ import RelativeTime from './relative-time'
...
+const nowElements = new Set()
+
+// Internal: Timer ID for `updateNowElements` interval.
+let updateNowElementsId
+
+// Internal: Install a timer to refresh all attached relative-time elements every
+// minute.
+function updateNowElements() {
+  clearTimeout(updateNowElementsId)
+  for (const el of nowElements) el.textContent = el.getFormattedDate() || ''
```
srt32, line 18: This is perhaps not related to this change, per se, but if I understand correctly, anytime a new `<relative-time>` element appears, we'll run `updateNowElements()` since it's called from ` connectedCallback()`. This seems wasteful since we'll go update all other elements on the page and just the new one. Am I understanding that correctly?

Maybe related, the parent runs https://github.com/github/time-elements/blob/4414cbacdb1e0cf95b7e99960897bb0c55736323/src/extended-time-element.ts#L16-L19 as part of `connectedCallback` so do we even need to worry about setting the initial value from this class?

keithamus: This is a valid point! We also have changed the semantics of this element because of this change!

src/relative-time-element.js
```diff
@@ -13,44 +28,13 @@ export default class RelativeTimeElement extends ExtendedTimeElement {
...
 
   connectedCallback() {
-    nowElements.push(this)
-
-    if (!updateNowElementsId) {
-      updateNowElements()
-      updateNowElementsId = setInterval(updateNowElements, 60 * 1000)
-    }
+    nowElements.add(this)
+    updateNowElements()
```
keithamus, line 32:
```suggestion
    if (!updateNowElementsId) updateNowElements()
```

src/relative-time-element.js
```diff
@@ -4,6 +4,21 @@ import RelativeTime from './relative-time'
...
+// to be updated every minute.
+const nowElements = new Set()
+
+// Internal: Timer ID for `updateNowElements` interval.
+let updateNowElementsId
+
+// Internal: Install a timer to refresh all attached relative-time elements every
+// minute.
+function updateNowElements() {
+  clearTimeout(updateNowElementsId)
```
stof, line 17: shouldn't this reset the updateNowElementsId variable too ? Otherwise, we could end up in a case where the timer won't run anymore:

1. add an element in the DOM. The `connectedCallback` will run the update and schedule a timer
2. remove the element from the DOM.
3. wait a bit for the timer to run.  `nowElements.size` is 0 so the timer is not rescheduled. but currently, `updateNowElementsId` keeps the old integer for the timer that just ran
4. add another element in the DOM. The `connectedCallback` will **not** run the update, because `updateNowElementsId` is already set (except that this is not set for a scheduled timer but for an old one). so the timer is never restarted

keithamus:
Closing for now. Will revisit when I have time.

PR #150, opened 2021-08-05
Update developer dependencies

koddsson:
Does what it says on the tin! No changes to production code.

Closes https://github.com/github/ui-frameworks/issues/118


dgreif (approved):
Looks great, thanks for keeping these updated! 🧹 ✨

PR #169, opened 2022-08-25
Add extensions to all imports

nwalters512:
In #168, I unfortunately forgot to add extensions to all the other files. This is necessary because their imports still appear in the generated `.d.ts` files.


dgreif (approved):
Good catch! I didn't get a chance to test it out yet, so I hadn't caught the missing extensions. Thanks for the update!

PR #170, opened 2022-08-25
Dependency updates and require extensions in imports

dgreif:
#168 and #169 added file extensions to imports, but it's easy to accidentally forget these in the future. As a safety net, I'd like to add the  `import/extensions` rule.

I also update all of our dependencies and tweaked the eslint rules to match existing code


keithamus (approved):
:+1:

PR #171, opened 2022-09-22
Update README.md

josepmartins:
- [x] Fix `time-until` output examples:
- [x] Add missing micro-format to `time-until`  

Shown in the example page: https://github.github.io/time-elements/examples/
Tested here: https://codepen.io/josepmartins/pen/abGOPVZ

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,11 @@ So, a relative date phrase is used for up to a month and then the actual date is
 
 You can use `time-until` to always display a relative date that's in the future. It operates much like `<relative-time>`, except in the reverse, with past events shown as `just now` and future events always showing as relative:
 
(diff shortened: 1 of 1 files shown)


koddsson:
Is the micro format not already documented? Or is it not clear that it's available to `<time-until>`?

https://github.com/github/time-elements#micro-format

josepmartins:
> Is the micro format not already documented? Or is it not clear that it's available to `<time-until>`?
> 
> https://github.com/github/time-elements#micro-format

That's a section of `micro` format on `time-ago` but not in `time-until`. I think it's not clear that both support that.

keithamus (approved):
:+1: LGTM

PR #178, opened 2022-10-19
switch from karma to web-test-runner

keithamus:
This switches us from `karma` to `@web/test-runner` which uses Playwright; it is slightly faster but also much easier to use and will allow us to migrate tests to TypeScript with little effort. It is also now our [recommended default test runner for custom elements](https://github.com/github/custom-element-boilerplate/blob/main/web-test-runner.config.js).

PR #179, opened 2022-10-19
Use esbuild

keithamus:
This switches us from Rollup to Esbuild to bundle the modules into a single file. We don't really _need_ to bundle, and maybe we can remove the bundling altogether in a major release, but using `esbuild` aligns us better to #178 which uses esbuild under the hood.

This also changes the way we bundle by separating the `dist/bundle.js` file as the `main`, allowing for upstream bundlers that use `module` keys to get the un-bundled code.


primer-css:

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@github/time-elements",
   "version": "3.1.2",
-  "main": "dist/index.js",
+  "main": "dist/bundle.js",
```
lukasoppermann, line 4: Just asking, this does not create issues for consumers how do a `require('@github/time-elements')` right? They would still get it, because this looks for whatever is `main` and not for `index.js`?

keithamus: `require()` will key off the `main` key, which defaults to `index.js` - but we've had the `main` key forever so this changes nothing in that regard.

More sophisticated bundlers will look at `module`.

lukasoppermann (approved):

PR #190, opened 2022-10-26
RelativeTimeElement: Progressive time update intervals

keithamus:
This takes the work from #122 and updates it in line with the changes to the codebase. (Subsequently, this closes #122)

When updating `nowElements`, figure out how far away in time the nearest one is to ensure timely updates:
If an element is within
  - one minute from now, update a second after last run
  - one hour from now - update a minute after last run
  - more than one hour from now - update hourly after last run

Switching to setTimeout ensures that with a lot of attached instances, where a run might take a while, we won't schedule the next run until after all elements have been updated.

Also fire a `relative-time-updated` event if `textContent` changes in case this needs to be acted on.

Additionally it "unobserves" elements that are no longer displaying relative time, meaning the amount of time elements to update can decrease if they're not relative, or have fallen from the threshold of displaying relative time.

It also adds an `update()` method to house a lot of this logic, and a new private class to handle the observation of updates, making the concrete class a little simpler to reason about.

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -4,6 +4,50 @@ import {makeFormatter, localeFromElement} from './utils.js'
 import {isDuration, withinDuration} from './duration.js'
 import {strftime} from './strftime.js'
 
+export class RelativeTimeUpdatedEvent extends Event {
+  constructor(public oldText: string, public newText: string, public oldTitle: string, public newTitle: string) {
+    super('relative-time-updated', {bubbles: true, composed: true})
+  }
+}
+
+function getUnitFactor(ms: number): number {
+  ms = Math.abs(Date.now() - ms)
+  if (ms < 60 * 1000) return 1000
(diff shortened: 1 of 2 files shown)


src/relative-time-element.ts
```diff
@@ -4,6 +4,50 @@ import {makeFormatter, localeFromElement} from './utils.js'
...
+    this.update()
+  }
+
+  unobserve(element: RelativeTimeElement) {
+    if (!this.elements.has(element)) return
+    this.elements.delete(element)
+    this.update()
+  }
+
+  timer: ReturnType<typeof setTimeout> = -1 as unknown as ReturnType<typeof setTimeout>
```
keithamus, line 35: TS really needs to fix its setTimeout signatures :sweat_smile:

primer-css:

simurai (approved):

PR #193, opened 2022-11-02
Switch element to defining content within ShadowDOM

keithamus:
This changes the `<relative-time>` element to use a ShadowRoot, instead of mutating the `textContent` directly. This is useful because it enables `<relative-time>` to mutate its own display without interfering with components or frameworks which assess or otherwise depend on the tree below `<relative-time>`; for example frameworks like React which maintain a virtual DOM, or code which leverages Mutation Observers to track changes to the tree.

The trade-off here is that it is not possible to assert the display of the date via `el.textContent`, as this would represent the light HTML tree (so likely the original date rendered server side). Consequently the tests need to change to accommodate this.

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -256,6 +256,11 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
     this.datetime = value?.toISOString() || ''
   }
 
+  constructor() {
+    super()
+    if (!this.shadowRoot) this.attachShadow({mode: 'open'})
+  }
+
   connectedCallback(): void {
     this.update()
   }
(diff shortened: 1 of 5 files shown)


src/relative-time-element.ts
```diff
@@ -256,6 +256,11 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
     this.datetime = value?.toISOString() || ''
   }
 
+  constructor() {
+    super()
+    if (!this.shadowRoot) this.attachShadow({mode: 'open'})
```
koddsson, line 261: I like the guard for declarative shadow doms! <3

koddsson (approved):
Looks great to me!

primer-css:

PR #194, opened 2022-11-03
Local time refactor

keithamus:
This PR introduces a `DateTimeFormat` ponyfill, which allows us to drop it in place of `Intl.DateTimeFormat`. This then allowed a larger refactor where we can drop parts of the codebase where we emulated `DateTimeFormat` behavior using strftime. Instead, the ponyfill does that and the `relative-time` and `local-time` code can be greatly simplified.

I think in the major version release, we should drop both the `DateTimeFormat` and `RelativeTime` ponyfills as these are very well supported in evergreen browsers. (`DateTimeFormat` works back to IE11 & Safari 10, while `RelativeTime` works back to Safari 14)


koddsson (approved):
Looks great!

PR #195, opened 2022-11-03
add tests for ponyfills

keithamus:
In #194 I forgot to add tests, which I had written. I've written some more now for both ponyfills so that they align with what the real Intl objects do (at least for `'en'` locale :smile:)


koddsson (commented):
Looks good!

PR #198, opened 2022-11-03
Make duration class

keithamus:
This refactors our `duration.ts` file to include a class which aligns closer to Temporal.Duration. In fact I even added `@js-temporal/polyfill` as a devdependency and ran some tests comparing our `Duration` class to the `Temporal.Duration` class.

This is useful because `threshold` can be set to a string that `Temporal.Duration.from` would take, and so the closer we align to that the better. Additionally, the closer our class is to `Temporal.Duration`, the easier it'll be to delete it one day.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "devDependencies": {
         "@custom-elements-manifest/analyzer": "^0.6.4",
         "@github/prettier-config": "0.0.4",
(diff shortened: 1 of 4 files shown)


primer-css:

simurai (approved):

koddsson (approved):

PR #200, opened 2022-11-07
avoid eagerly calling this.update()

keithamus:
`update()` is getting a bit too eagerly called in the latest version, which is causing many time elements on the page to stack overflow. This mitigates that in a few ways:

 - Ensure if the `attributeChangedCallback` is called with the same value, that it doesn't call `update`.
 - Ensure that if the `title` attribute changes, it doesn't lead to an `update`.
 - Ensure that when an element gets `unobserve`d, that it doesn't trigger an update of all elements (as there's no point in updating eagerly here - the current `setTimeout` could already be eager if the element being unobserved is the latest element).
 - Ensure that when an element _does_ get `observe`d, that it doesn't immediately trigger an update. Rather, it determines if it _should_ update sooner than the next scheduled update, and if so it'll reschedule the update to occur sooner.

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -28,17 +28,26 @@ function getUnitFactor(ms: number): number {
 
 const dateObserver = new (class {
   elements: Set<RelativeTimeElement> = new Set()
+  time = 0
 
   observe(element: RelativeTimeElement) {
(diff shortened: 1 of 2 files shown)


koddsson (approved):
Looks good!

primer-css:

PR #201, opened 2022-11-07
fix bug in Intl detection

keithamus:
In testing this in various browsers I had noticed when it needs to fall back to the ponyfills, it won't correctly do so due to a bug (I think from refactoring) the feature detection. This fixes that.


koddsson:
Very cool @keithamus!

PR #210, opened 2022-11-23
add `define` export for custom element registration

keithamus:
Following a discussion with @koddsson - this PR attempts to create a pattern of module definitions within our custom elements.

Effectively, all custom elements can be imported via their bare specifier (so in this case `@github/relative-time-element`). This _will not import the custom element definition_ (though it does here, to avoid breaking changes). Instead one will have to `import '@github/relative-time-element/define'`. This will then import the code which defines the element.

In addition, for this element (at least for now - though these will be removed in a breaking release) the following imports are also enabled:

- `'@github/time-elements` - all classes
- `'@github/time-elements/define` - all classes and their custom element definitions
- `'@github/time-elements/relative-time` - just the relative time element
- `'@github/time-elements/relative-time/define` - just the relative time element and its custom element definition
- `'@github/time-elements/local-time` - just the local time element
- `'@github/time-elements/local-time/define` - just the local time element and its custom element definition
- `'@github/time-elements/time-ago` - just the time-ago element
- `'@github/time-elements/time-ago/define` - just the time-ago element and its custom element definition
- `'@github/time-elements/time-until` - just the time-until element
- `'@github/time-elements/time-until/define` - just the time-until element and its custom element definition

The plan is to remove a lot of these elements in v4, so the following list will be true for v4:

- `'@github/time-elements` - the `relative-time` element
- `'@github/time-elements/define` - the relative time element and its custom element definition

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,18 @@
     "dist",
     "custom-elements.json"
   ],
+  "exports": {
+    ".": "./dist/index.js",
+    "./define": "./dist/index.js",
+    "./relative-time": "./dist/relative-time-element.js",
+    "./relative-time/define": "./dist/relative-time-element-define.js",
+    "./local-time": "./dist/local-time-element.js",
+    "./local-time/define": "./dist/local-time-element-define.js",
+    "./time-ago": "./dist/time-ago-element.js",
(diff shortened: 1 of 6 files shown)


primer-css:

PR #213, opened 2022-11-23
add JSX IntrinsicElement types

### LINKED ISSUE #163: Add better jsx support
Properly add the various elements to the JSX intrinsic space similar to:

```type CustomElement<T> = Partial<T & DOMAttributes<T> & { children: any }>;

declare global {
  namespace JSX {
    interface IntrinsicElements {
      ["time-ago"]: CustomElement<TimeAgoElement>;
    }
  }
}```

### PR DESCRIPTION
**Pull request by keithamus**
This adds the `JSX.IntrinsicElement` types to the `define` files. Closes #163

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,12 @@
   },
   "overrides": [
     {
+      "files": "src/*-define.ts",
+      "rules": {
+        "@typescript-eslint/no-namespace": "off"
+      }
+    },
+    {
       "files": "test/**/*.js",
       "rules": {
         "github/unescaped-html-literal": "off",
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by JoshBowdenConcepts — APPROVED** [review]
Approving for @github/primer-reviewers

**Comment by koddsson** [comment]
Are we planning to auto-generate these in the future? I think that could be cool with the Custom Element Manifest if possible. Then it should be trivial to add other bindings for other frameworks as well and we don't have to manually maintain these bindings.

**Comment by keithamus** [comment]
Experimenting with this a bit, this is untenable. React requires a type that includes references to the React library itself, which we're not prepared to bring into this custom element.

```ts
declare global {
  namespace JSX {
    interface IntrinsicElements {
      'relative-time': React.DetailedHTMLProps<React.HTMLAttributes<RelativeTimeElement>, RelativeTimeElement>
    }
  }
}
```

This should be defined in consuming libraries if they want to use this in React.

PR #214, opened 2022-11-25
add JSX IntrinsicElement types

keithamus:
Okay after a bit of a battle I managed to conjur the right typescript incantations which enables us to have operational typescript types, while also not _requiring_ JSX or React types to be installed.

This is a re-opening of https://github.com/github/relative-time-element/pull/213. Something @koddsson asked in hat PR but I didn't answer:

> Are we planning to auto-generate these in the future? I think that could be cool with the Custom Element Manifest if possible. Then it should be trivial to add other bindings for other frameworks as well and we don't have to manually maintain these bindings.

We could definitely take a look at this but in other frameworks this isn't really an issue as they either support custom elements well enough to not require this additional support scaffolding (perhaps by using the `HTMLElementTagNameMap` type) or they use the `JSX.IntrinsicElements` definition anyway. Which is all to say I think we won't need to support any other frameworks.

### DIFF
--- a/src/local-time-element-define.ts
+++ b/src/local-time-element-define.ts
@@ -13,12 +13,20 @@ try {
   }
 }
 
+type JSXBaseElement = JSX.IntrinsicElements extends {span: unknown}
+  ? JSX.IntrinsicElements['span']
+  : Record<string, unknown>
 declare global {
   interface Window {
     LocalTimeElement: typeof LocalTimeElement
   }
   interface HTMLElementTagNameMap {
     'local-time': LocalTimeElement
+  }
+  namespace JSX {
+    interface IntrinsicElements {
+      ['local-time']: JSXBaseElement & Partial<Omit<LocalTimeElement, keyof HTMLElement>>
+    }
   }
 }
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,12 @@
   },
   "overrides": [
     {
+      "files": "src/*-define.ts",
+      "rules": {
+        "@typescript-eslint/no-namespace": "off"
+      }
+    },
+    {
       "files": "test/**/*.js",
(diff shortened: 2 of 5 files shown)


primer-css:

simurai (approved):

src/local-time-element-define.ts
```diff
@@ -13,13 +13,21 @@ try {
   }
 }
 
+type JSXBaseElement = JSX.IntrinsicElements extends {span: unknown}
+  ? JSX.IntrinsicElements['span']
+  : Record<string, unknown>
```
koddsson, line 18: I'm having a really hard time reading what this does 😅 

JSXBaseElement is a the JSX element `span` _if_ JSX elements extend `{span: unknwon}`? Otherwise it's just a record of `string, unknown`?

koddsson (approved):
Looks good to me. Sounds like this won't need a lot of future maintance so I'm fine with the extra complexity.

Maybe some comments explaining how the JSX element types work and how we are using them could be good?

src/local-time-element-define.ts
keithamus: Yeah this is a guard that essentially checks if `JSX.InstrinsicElements.span` exists. If it doesn't exist, as you say, then the type is `Record<string, unknown>`.

In other words, if a user installs this webcomponent and doesn't have `@types/react` then this will create:

```ts
JSX.InstrinsicElements['relative-time'] = Record<string, unknown> & Partial<Omit<RelativeTimeElement, keyof HTMLElement>>
```

If they _do_ have `@types/react` then this will create:

```ts
JSX.InstrinsicElements['relative-time'] = JSX.InstrinsicElements.span & Partial<Omit<RelativeTimeElement, keyof HTMLElement>>
```

PR #217, opened 2022-11-28
Warn on use of deprecated elements, functions and properties

keithamus:
This PR is a BREAKING CHANGE, in which we `console.warn` on all deprecated features, which we plan to remove in the next version, `v4.0.0`.

The following features are to be deprecated in `v4.0.0`:

 - The `local-time` element will be removed.
 - The `time-ago` element will be removed.
 - The `time-until` element will be removed.
 - `getFormattedTitle` will be private. 
 - `getFormattedDate` will be private. 
 - `format` will no longer accept a `strftime` compatible format. Instead, any value other than `'micro'`, `'elapsed'` will revert to the string `'auto'`
 - the `Intl.DateTimeFormat` constructor will not be ponyfilled, and so in environments without this constructor, the 
element will fall back to provided text.
 - the `Intl.RelativeTimeFormat` constructor will not be ponyfilled, and so in environments without this constructor, the element will fall back to provided text.

This PR **maintains compatibility with Version 3** but will `console.warn` on the use of any of these deprecated features. Additionally the typescript definitions have changed which means integrators will need to alter their code for typescript.

The plan is to release this as `v4.0.0-pre`, which then follows up with a `v4.0.0` proper, which removes all of the above features. This will allow integrators to install `v4.0.0-pre` and ship it to production in a backwards-compatible way with `v3.x`, while refactoring out any uses of deprecated APIs, before finally installing `v4.0.0`

### DIFF
--- a/src/datetimeformat-ponyfill.ts
+++ b/src/datetimeformat-ponyfill.ts
@@ -3,6 +3,10 @@ export class DateTimeFormat implements Intl.DateTimeFormat {
   #options: Intl.ResolvedDateTimeFormatOptions
 
   constructor(locale: string, options: Intl.DateTimeFormatOptions) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `time-elements v5.0.0 will no longer ship with DateTimeFormat ponyfill. It must be polyfilled for continued support in older browsers`
+    )
     this.#options = {
       locale: 'en',
       calendar: 'gregory',
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -127,7 +127,7 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
     }).format(date)
   }
 
-  getFormattedTitle(): string | undefined {
+  private getFormattedTitle(): string | undefined {
     // eslint-disable-next-line no-console
     console.warn(`Calling getFormattedTitle is deprecated and will be removed in v5.0.0`)
     return this.#getFormattedTitle()
@@ -174,7 +174,7 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
     return `${this.prefix} ${formatter.format(date)}`.trim()
   }
 
-  getFormattedDate(now = new Date()): string | undefined {
+  private getFormattedDate(now = new Date()): string | undefined {
     // eslint-disable-next-line no-console
     console.warn(`Calling getFormattedTitle is deprecated and will be removed in v5.0.0`)
     return this.#getFormattedDate(now)
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -12,7 +12,7 @@ const DateTimeFormat = supportsIntlDatetime ? Intl.DateTimeFormat : DateTimeForm
 const supportsIntlRelativeTime = typeof Intl !== 'undefined' && 'RelativeTimeFormat' in Intl
 const RelativeTimeFormat = supportsIntlRelativeTime ? Intl.RelativeTimeFormat : RelativeTimeFormatPonyfill
 
-export type Format = 'auto' | 'micro' | 'elapsed' | string
+export type Format = 'auto' | 'micro' | 'elapsed'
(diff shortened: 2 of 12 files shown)


src/datetimeformat-ponyfill.ts
```diff
@@ -3,6 +3,10 @@ export class DateTimeFormat implements Intl.DateTimeFormat {
   #options: Intl.ResolvedDateTimeFormatOptions
 
   constructor(locale: string, options: Intl.DateTimeFormatOptions) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `time-elements v5.0.0 will no longer ship with DateTimeFormat ponyfill. It must be polyfilled for continued support in older browsers`
```
lukasoppermann, line 8: This could use a normal string and not a template string.

src/relative-time-element.ts
```diff
@@ -127,7 +127,13 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
     }).format(date)
   }
 
-  getFormattedDate(now = new Date()): string | undefined {
+  private getFormattedTitle(): string | undefined {
+    // eslint-disable-next-line no-console
+    console.warn(`Calling getFormattedTitle is deprecated and will be removed in v5.0.0`)
```
lukasoppermann, line 132: Same here, normal string instead of template string would be fine.

src/relative-time-element.ts
```diff
@@ -168,6 +174,12 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
     return `${this.prefix} ${formatter.format(date)}`.trim()
   }
 
+  private getFormattedDate(now = new Date()): string | undefined {
+    // eslint-disable-next-line no-console
+    console.warn(`Calling getFormattedTitle is deprecated and will be removed in v5.0.0`)
```
lukasoppermann, line 179: Same here, normal string instead of template string would be fine.

src/relative-time-element.ts
```diff
@@ -301,11 +313,23 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
     const format = this.getAttribute('format')
     if (format === 'micro') return 'micro'
     if (format === 'elapsed') return 'elapsed'
-    if (format && format.includes('%')) return format
+    if (format && format.includes('%')) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `srftime formatting is deprecated and will be removed in v5.0.0. stftime formats will default to 'auto'`
```
lukasoppermann, line 319: Same here, normal string instead of template string would be fine.

src/relative-time-element.ts
```diff
@@ -301,11 +313,23 @@ export default class RelativeTimeElement extends HTMLElement implements Intl.Dat
...
+      return format as unknown as Format
+    }
     return 'auto'
   }
 
   set format(value: Format) {
+    if (value && value.includes('%')) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `srftime formatting is deprecated and will be removed in v5.0.0. stftime formats will default to 'auto'`
```
lukasoppermann, line 330: Same here, normal string instead of template string would be fine.

src/relative-time-ponyfill.ts
```diff
@@ -1,4 +1,11 @@
 export class RelativeTimeFormat implements Intl.RelativeTimeFormat {
+  constructor() {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `time-elements v5.0.0 will no longer ship with DateTimeFormat ponyfill. It must be polyfilled for continued support in older browsers`
```
lukasoppermann, line 5: Same here, normal string instead of template string would be fine.

lukasoppermann (approved):
Seems legit.

koddsson (approved):
Looks good to me!

Regarding the template strings. I don't see why they can't be template strings. If we want to discourage template strings where we aren't interpolating strings and data then we should have a lint rule for that and let CI catch those violations.

keithamus:
Given they're temporary I'd sooner merge this then delete :smile:

PR #225, opened 2022-12-07
fix: dasherize formatStyle in observedAttributes

lyricwulf:
observed attributes should be dasherized, not camel 🐶


keithamus (approved):
🤦‍♂️ yes of course, thank you!

PR #226, opened 2022-12-08
Refactor micro time to use a rounded duration

keithamus:
This is a big refactor of the `micro` format, to remove our rather strange rounding logic and instead use Duration with some logic which is a bit easier to reason about (and potentially easier to delete).

The main benefit here is that:

1. `micro` format is now like `elapsed` but with some extra logic. This pairs down a lot of the complexity we had with the format.
2. The `micro` format is now somewhat easier to deprecate, which is our intent with v5.
3. The format relies more on our `Duration` object which is a ponyfill for `Temporal.Duration`, and so one day can be deleted (perhaps a v6 version :wink:)
4. This lays ground work for us to better refactor the `timeAgo` and `timeUntil` functions which just need to be wrangled to use the same rounding logic.

It should be noted that this _does_ change some tests! Mostly around the ways with which the micro format gets displayed. Units are rounded up in what might be considered more expected ways. Week units are introduced (so `4d`, `5d`, `6d`, `7d` will all now become `1w`). Days are rounded to months, so `31d` is `1m`, and days are now rounded to years (so `364d` becomes `1y`).


primer-css:

PR #245, opened 2023-02-01
update rounding rules for months/years in Duration

keithamus:
Currently the main release of relative time uses a 90% rounding rule for dates.

Relative times are colloquial and so every user interprets them in subtly different ways. There are a few techniques we have used in the past:

 - We go on the literal date unit. So at 12:01AM a post from 2 minutes prior would say "a day ago". A post on the 31st December, when read on 1st Jan, would say "last year", which feels inaccurate.
 - We round units to some amount, so if the delta is 2 minutes we say "2 minutes ago" until minutes can be rounded up to hours at which point we can say "1 hour ago". We can pick the threshold for this...
   - A threshold of 50% would mean something 30 minutes ago would be "1 hour ago", and something 6 months ago would say "last year". Having a low threshold for larger durations seems inaccurate.
   - A threshold of 90% would mean something 54 minutes ago would be "1 hour ago", something 11 months ago would be "last year".  The downside to this is that occasionally you'll get rounding such as something being 20 months old saying "last year". Having a high threshold seems inaccurate.

Right now the 90% threshold is our latest iteration but it has problems with large units of time, chiefly months and years. At the time of writing, for example, dates from April 2021 will be marked as "last year" despite being much closer to "2 years ago".

This PR changes the logic _just for months and years_ to use a similar mechanism to that described in [`Temporal.Duration.round`](https://tc39.es/proposal-temporal/docs/duration.html#round). We take the `relativeTo` which is a date, and we execute calendar math on the date object - working out the delta of the current month and rounding based on calendar year instead of a rounding of months.

This means for dates that have a delta of months or years, they will perform calendar operations, and those dates will be relative to the calendar year. In other words a date from 2022 read in 2023 will only ever read as "last year" and not "2 years ago", meanwhile a date from 2021 read in 2023 will never say "last year" and will always say "2 years ago".

### DIFF
--- a/src/duration.ts
+++ b/src/duration.ts
@@ -119,8 +119,14 @@ export function elapsedTime(date: Date, precision: Unit = 'second', now = Date.n
   )
 }
 
-export function roundToSingleUnit(duration: Duration): Duration {
+interface RoundingOpts {
+  relativeTo: Date | number
+}
+
+export function roundToSingleUnit(duration: Duration, {relativeTo = Date.now()}: Partial<RoundingOpts> = {}): Duration {
+  relativeTo = new Date(relativeTo)
   if (duration.blank) return duration
+  const sign = duration.sign
   let years = Math.abs(duration.years)
   let months = Math.abs(duration.months)
   let weeks = Math.abs(duration.weeks)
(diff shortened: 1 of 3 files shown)


primer-css:

rezrah (approved):
Looks good, and a nice improvement. Thanks @keithamus.

I couldn't get all tests to pass locally, but CI verified no regressions.

PR #257, opened 2023-04-18
add define static method

keithamus:
This makes a few small tweaks to how we define elements. 

It adds a `define()` static member to the `RelativeTimeElement` class, which allows changing of tag/registry. It calls that method in the `relative-time-element-define` file.

This also exports the class as both named and default exports. We may drop the default export at a future date.


primer-css:

Issue #262, opened 2023-07-11
Stray days added to duration


TymekDev:
Hey, I am having an issue where the relative time between `2022-07-11T09:00:00+02:00` and `2023-07-11T09:00:00+02:00` is `1 year, 5 days` instead of `1 year`. Similarly `2023-05-11T09:00:00+02:00` and `2023-07-11T09:00:00+02:00` are `2 months, 1 day` where I'd expect it to be  just `2 months`. As I shorten or lengthen the difference between datetimes the days decrease or increase respectively.

I can think of possibilities at calculation level where these come from, however I find the human-readable duration confusing.

francinelucca:
👋🏽 validated failing test cases are valid https://github.com/github/relative-time-element/pull/263 and spent some time during FR shift [chugging at a solution ](https://github.com/github/relative-time-element/pull/376). Unfortunately the fix is not straight forward as there's a lot of considerations/decisions to be made here. Since this is a bigger effort that falls outside the scope of FR, adding to team inbox so we can discuss at triage.

PR #263, opened 2023-07-11
add test cases for superfluous days

TymekDev:
This PR adds cases for whole years and whole months to not have extra days, e.g. May (`2023-05-11T00:00:00.00`) to July (`2023-07-11T00:00:00.00`) being `2 months` instead of current `2 months, 1 day`.

Related issue: #262


lukasoppermann:
I'd love for @keithamus to approve this as he is working on it. Fine from primer FR side

keithamus:
Thanks, I'm looking into fixing this but I have higher priorities for now. A fix will come though!

joshblack:
Sorry for the delay on my end, adding this to the first responder board to pick up 👀 @keithamus just wanted to check-in, what are the next steps for this?

camertron:
I looked into this today and have become thoroughly confused. Are you specifying any `precision` argument anywhere?

If I try copy/pasting your test cases into example/index.html, it always gives me something like "1 year, 6 days, 4 minutes, 24 seconds." I don't understand why the tests fail the way they do when no `precision` is specified.

TymekDev:
@camertron according to the README the precision defaults to `second`, so I'd assume that's how it works. I don't think it changes anything in this context, though.

I just put this into an HTML (using v4.4.4 via npm):

```html
<relative-time format="duration" datetime="1944-12-13"></relative-time>
```

It showed "81 years, 2 months, 22 hours, 21 minutes, 45 seconds" for me. Even if I set the precision as high as `year`, then that would still be wrong—it should be 80 years. In other words, given a long enough duration, all precision levels are affected.

francinelucca:
Thanks for opening this PR @TymekDev! sorry for lack of activity here. I've taken a look at the tests and looks like you're right, the code needs to be corrected. I don't want to get failing tests into main so I'm hoping we can land on a fix instead and then pull these in. I've dropped a comment in https://github.com/github/relative-time-element/issues/262#issuecomment-5441285179, hopefully we can work on this soon. Will leave this PR open until then.

francinelucca (approved):

PR #275, opened 2024-03-07
Fix default value of `lang`

silverwind:
As per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang):

> The default value of `lang` is `unknown`

I'm pretty sure the previous value `default` was actually invalid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ For dates outside of the specified `threshold`, the formatting of the date can b
 
 ##### lang
 
-Lang is a [built-in global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang). Relative Time will use this to provide an applicable language to the `Intl` APIs. If the individual element does not have a `lang` attribute then it will traverse upwards in the tree to find the closest element that does, or default the lang to `en`.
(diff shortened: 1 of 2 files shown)


mperrotti (approved):

silverwind:
Upon further investigation, I think MDN is in error. The "unknown" word only appears in old [W3C specs](https://www.w3.org/TR/html401/struct/dirlang.html#language-info), but as literal text:

> The default value of this attribute is unknown

The [WHATWG spec](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes) recommends the empty string as the default.

> Its value must be a valid BCP 47 language tag, or the empty string. Setting the attribute to the empty string indicates that the primary language is unknown

keithamus:
Yeah `default` was carefully chosen as I think it doesn't reject. In reality I think this is somewhat undefined behavior for this element as anywhere we expect it to be used should have at _minimum_ a `lang` attribute on the HTML element.

silverwind:
I think for `lang` being a HTML attribute, empty string makes most sense to me. Intl APIs like `Intl.DateTimeFormat` not accepting empty string is a interop issue, but it also raises the question whether a `lang` is equivalent to a `locale` or whether they are different concepts.

As far as I understand `lang=''` is equivalent to `locales=[]`, both result in "current locale" behaviour.
 
I submitted https://github.com/mdn/content/pull/32604 to correct MDN.

PR #277, opened 2024-03-14
Run `npx eslint . --ext .js,.ts --fix`

silverwind:
`npm run build` refused to proceed building without this.

I think the issue is because some eslint module is not properly locked down to a particular version. It would be even better to not have linting as a dependency to the build.

### DIFF
--- a/src/duration-format-ponyfill.ts
+++ b/src/duration-format-ponyfill.ts
@@ -108,8 +108,8 @@ export default class DurationFormat {
         unitStyle === '2-digit'
           ? twoDigitFormatOptions
           : unitStyle === 'numeric'
-          ? {}
-          : {style: 'unit', unit: nfUnit, unitDisplay: unitStyle}
+            ? {}
+            : {style: 'unit', unit: nfUnit, unitDisplay: unitStyle}
       list.push(new Intl.NumberFormat(locale, nfOpts).format(value))
     }
     return new ListFormat(locale, {
--- a/src/duration.ts
+++ b/src/duration.ts
@@ -2,7 +2,7 @@ import DurationFormat from './duration-format-ponyfill.js'
 import type {DurationFormatOptions} from './duration-format-ponyfill.js'
 const durationRe = /^[-+]?P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?$/
 export const unitNames = ['year', 'month', 'week', 'day', 'hour', 'minute', 'second', 'millisecond'] as const
-export type Unit = typeof unitNames[number]
+export type Unit = (typeof unitNames)[number]
 
 export const isDuration = (str: string) => durationRe.test(str)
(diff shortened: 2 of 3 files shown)


silverwind:
Hmm, the CI now suggests the inverse. There must be something wrong with my setup I suppose.

silverwind:
Nope, still present on fresh clone. Very strange:

```
$ npm i
added 583 packages in 34s
> @github/relative-time-element@0.0.0-development prebuild
> npm run clean && npm run lint && mkdir dist


> @github/relative-time-element@0.0.0-development clean
> rm -rf dist


> @github/relative-time-element@0.0.0-development lint
> eslint . --ext .js,.ts && tsc --noEmit


/Users/silverwind/git/relative-time-element/src/duration-format-ponyfill.ts
  111:11  error  Insert `··`  prettier/prettier
  112:1   error  Insert `··`  prettier/prettier

/Users/silverwind/git/relative-time-element/src/duration.ts
  5:20  error  Replace `typeof·unitNames` with `(typeof·unitNames)`  prettier/prettier

/Users/silverwind/git/relative-time-element/src/relative-time-element.ts
   2:67  error  Replace `typeof·window` with `(typeof·window)`                                                                                                                                                                                           prettier/prettier
  14:15  error  Replace `public·oldText:·string,·public·newText:·string,·public·oldTitle:·string,·public·newTitle:·string` with `⏎····public·oldText:·string,⏎····public·newText:·string,⏎····public·oldTitle:·string,⏎····public·newTitle:·string,⏎··`  prettier/prettier

✖ 5 problems (5 errors, 0 warnings)
  5 errors and 0 warnings potentially fixable with the `--fix` option.
```

keithamus:
What is your `git rev-parse HEAD`?

silverwind:
`1a42a317372e09b08fe430fe36f2598d2039bbfa`, tip of `main` branch. Also tried another machine, same result.

keithamus:
I'm unable to repro. I wonder if you have one of these tools installed globally which is overriding the local version or something? Whats the output of `npm ls -g prettier` or `npm ls -g eslint`?

silverwind:
Found the problem. My `~/.npmrc` contains [`lockfile-version=3`](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json#lockfileversion), but this package has `lockfileVersion=2`, which is a legacy format that contains two entries for each module, while version 3 only contains one.

Checking whether I can remove this without causing npm to write legacy lockfiles...

silverwind:
Yeah, it seems to still write v3 lockfiles into new projects, so I'll keep that. Sorry for the trouble. The problem could have been avoided by declaring `lockfileVersion` in `.npmrc` in the project, but I'm not convinced this is really worth the effort.

silverwind:
After testing https://github.com/github/relative-time-element/pull/278, it turns out I made an error (I had moved away the `.npmrc` and was not cleanly reinstalling node_modules). The actual cause in my `~/.npmrc` was [`package-lock=false`](https://docs.npmjs.com/cli/v10/using-npm/config#package-lock) which made npm ignore the lockfile entirely. I'm suprised this option hadn't bitten me sooner.

PR #278, opened 2024-03-15
update lockfile to v3

keithamus:
@silverwind

refs #277


silverwind (approved):
LGTM, even thought it was not the direct cause (see https://github.com/github/relative-time-element/pull/277#issuecomment-1999461631).

If npm v9 (which comes with node 18) can be required, this will work. If older node versions are to be supported, I recommend staying on lockfile v2.

PR #280, opened 2024-03-27
Express "0 precision-units away" appropriately for the precision.

### LINKED ISSUE #279: precision=day should return "today" instead of "now"
https://jsfiddle.net/oabcw7mz/ has an example of trying to render (now - 23 hours) with a precision of days, which should show "today" but actually shows "now".

I think the fix will involve making `getRelativeTimeUnit` https://github.com/github/relative-time-element/blob/a0483cee69830a46e16c2bb8969b601988a4e8e3/src/duration.ts#L206-L218

return the `precision` unit for empty durations.

jyasskin:
This turns "now" into "today", "this month", "this hour", etc.

Fixes #279.

Whether "this minute" is better than "now" is debatable, but I feel like that debate should happen in the CLDR data that drives `Intl.RelativeTimeFormat.format(0, 'minute')` rather than in this library.

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -180,7 +180,7 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     if (tense === 'past' && duration.sign !== -1) duration = emptyDuration
     const [int, unit] = getRelativeTimeUnit(duration)
     if (unit === 'second' && int < 10) {
-      return relativeFormat.format(0, 'second')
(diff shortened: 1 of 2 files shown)


keithamus (approved):
Nice work!

keithamus:
> Whether "this minute" is better than "now" is debatable, but I feel like that debate should happen in the CLDR data that drives `Intl.RelativeTimeFormat.format(0, 'minute')` rather than in this library.

Totally agree.

PR #283, opened 2024-05-29
Adds flag to optionally hide title attribute

lindseywild:
Related to https://github.com/github/accessibility-audits/issues/5633 & https://github.com/github/accessibility-audits/issues/5524.

Adds a flag to optionally hide the inaccessible `title` attribute. We will use this in the Primer components so we can add an accessible Tooltip element instead for interactive elements.

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -107,6 +107,7 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
       'precision',
       'format',
       'format-style',
+      'hide-title',
       'datetime',
       'lang',
       'title',
@@ -382,6 +383,10 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     this.setAttribute('format-style', value)
   }
 
+  get hideTitle(): boolean {
+    return this.getAttribute('hide-title') === 'true'
+  }
+
   get datetime() {
     return this.getAttribute('datetime') || ''
   }
@@ -431,7 +436,7 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
       return
     }
     const now = Date.now()
-    if (!this.#customTitle) {
+    if (!this.#customTitle && !this.hideTitle) {
       newTitle = this.#getFormattedTitle(date) || ''
       if (newTitle) this.setAttribute('title', newTitle)
     }
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -387,7 +387,7 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     return this.hasAttribute('no-title')
   }
 
-  set noTitle(value: boolean | null) {
+  set noTitle(value: boolean | undefined) {
     this.setAttribute('no-title', value?.toString() || '')
   }
 
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -384,7 +384,11 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
   }
 
   get noTitle(): boolean {
-    return this.hasAttribute('no-title') && this.getAttribute('no-title') === 'true'
+    return this.hasAttribute('no-title')
+  }
+
+  set noTitle(value: boolean | null) {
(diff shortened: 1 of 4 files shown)


src/relative-time-element.ts
```diff
@@ -382,6 +383,10 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     this.setAttribute('format-style', value)
   }
 
+  get hideTitle(): boolean {
+    return this.getAttribute('hide-title') === 'true'
```
keithamus, line 387: This should probably use hasAttribute

src/relative-time-element.ts
```diff
@@ -107,6 +107,7 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
       'precision',
       'format',
       'format-style',
+      'hide-title',
```
keithamus, line 110: I’d recommend `no-title`.

src/relative-time-element.ts
```diff
@@ -382,6 +383,10 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     this.setAttribute('format-style', value)
   }
 
+  get noTitle(): boolean {
+    return this.hasAttribute('no-title') && this.getAttribute('no-title') === 'true'
+  }
```
smockle, line 388: Non-blocking:
- Is there any reason to make this private?
- Is there any reason to add a setter (to support e.g. `relativeTime.noTitle = false`)?

(I can’t think of any good reasons.)

smockle (approved):

src/relative-time-element.ts
```diff
@@ -382,6 +383,10 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     this.setAttribute('format-style', value)
   }
 
+  get noTitle(): boolean {
+    return this.hasAttribute('no-title') && this.getAttribute('no-title') === 'true'
```
keithamus, line 387: Checking for `'true'` isn’t super idiomatic. I’d prefer us to check just for the presence of the attribute, using hasAttribute (so no getAttribute call). Then the markup can be `<relative-time no-title>`. I agree a setter might be useful here too, especially for React

lindseywild: Ok that makes more sense! So we would just pass in `no-title` with no value?

keithamus: Yup!

src/relative-time-element.ts
```diff
@@ -382,6 +383,14 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     this.setAttribute('format-style', value)
   }
 
+  get noTitle(): boolean {
+    return this.hasAttribute('no-title')
+  }
+
+  set noTitle(value: boolean | null) {
+    this.setAttribute('no-title', value?.toString() || '')
```
keithamus, line 391: This should probably use `toggleAttribute` which will be a bit simpler.

lindseywild: I had to pass in a param into the setter even though it's unused... Not sure how else to get around that as I believe it's required 🤔 Maybe it's ok?

src/relative-time-element.ts
```diff
@@ -382,6 +383,14 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     this.setAttribute('format-style', value)
   }
 
+  get noTitle(): boolean {
+    return this.hasAttribute('no-title')
+  }
+
+  set noTitle(value: boolean | undefined) {
+    this.toggleAttribute('no-title')
```
smockle, line 391: Consider providing a [`force` parameter](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute#force):

```suggestion
    this.toggleAttribute('no-title', value)
```

Otherwise, we’ll get surprising behavior like:
1. Initial state: `<relative-time no-title="true" />`.
2. Then, run `relativeTimeElement.noTitle = true`
3. New state: `<relative-time />`.

lindseywild: Ah, another thing I didn't know existed! Thank you!

keithamus (approved):
LGTM :+1:

PR #290, opened 2024-08-23
ensure #updating doesnt hang updates

### LINKED ISSUE #289: relative-time doesn't render anything without having an initial datetime value
I spoke a bit with @keithamus about this. I have a `hidden` `<relative-time>` element on the page that I want to programmatically reveal after I've set its `datetime` attribute. I learned that if `datetime` is omitted or left blank, then when I _do_ `setAttribute('datetime'`, nothing appears for the element on the page, and the `<relative-time>` element has a size of 0x0. If I manually call `.update()` on the element via my browser console, the '1 minute ago' or whatever timestamp appears as expected. 👍 

I have a workaround in that I can set `datetime` to the current time's ISO8601 string in the static page HTML. Then I found I can programmatically change the `datetime` attribute value and the updated timestamp appears like I want.

My use case is that I have a `<relative-time>` element that I want to populate once I know the right timestamp. I'm getting a timestamp from localstorage, so I have to check if the value exists and, if so, update `datetime` on the relative-time and display it.

keithamus:
@cheshire137 exposed an interesting timing flaw in #289. If you set a null/empty datetime the `#updating` Promise (which tries to batch updates) can be left as a resolved promise due to an early return in `update()`.

The last line of `update()` sets `#updating` to false:

https://github.com/github/relative-time-element/blob/15ca61e1992f39f242aa4de0b35c14ce76062990/src/relative-time-element.ts#L475-L476

However if `update()` is called while the `date` is `null` then it hits this early return:

https://github.com/github/relative-time-element/blob/15ca61e1992f39f242aa4de0b35c14ce76062990/src/relative-time-element.ts#L438-L441

That early return never sets `#updating = false`, and so when the date is set again we skip a call to `update()` as it looks like we're in an update batch already:

https://github.com/github/relative-time-element/blob/15ca61e1992f39f242aa4de0b35c14ce76062990/src/relative-time-element.ts#L425-L430

The solution is quite simple: update should always set `#updating = false` when returning. So one could assume the following diff would be a good patch:

```diff
    if (typeof Intl === 'undefined' || !Intl.DateTimeFormat || !date) {
      this.#renderRoot.textContent = oldText
+     this.#updating = false
      return
    }
```

While this would solve the problem it doesn't make for very maintainable code as each new early return would also need to add this or reintroduce the bug. So instead we should colocate all `#updating` assignments, which should have happened in the first place. Hence the patch is:

```diff
    if (!this.#updating && !(attrName === 'title' && this.#customTitle)) {
      this.#updating = (async () => {
        await Promise.resolve()
        this.update()
+       this.#updating = false
      })()
    }
```

Fixes #289.

### DIFF
--- a/test/relative-time.js
+++ b/test/relative-time.js
@@ -70,6 +70,15 @@ suite('relative-time', function () {
     assert.equal(counter, 1)
   })
 
+  test('calls update even after nullish datetime', async () => {
+    const el = document.createElement('relative-time')
+    el.setAttribute('datetime', '')
+    await new Promise(resolve => setTimeout(resolve, 10))
+    el.setAttribute('datetime', new Date().toISOString())
+    await Promise.resolve()
+    assert(el.shadowRoot.textContent.length > 0, 'should have set time, but textContent is empty')
+  })
+
   test('sets title back to default if removed', async () => {
     const el = document.createElement('relative-time')
     el.setAttribute('datetime', new Date().toISOString())
--- a/examples/index.html
+++ b/examples/index.html
@@ -193,6 +193,13 @@
     </relative-time>
   </p>
 
+  <p>
+    Lazily added datetime via setAttribute
+    <relative-time id="lazy" tense="past">
+      This will display in one second.
+    </relative-time>
+  </p>
+
   <!-- <script type="module" src="../dist/index.js"></script> -->
   <script type="module" src="https://unpkg.com/@github/relative-time-element@latest/dist/bundle.js"></script>
   <script>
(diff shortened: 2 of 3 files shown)


lukasoppermann (approved):

test/relative-time.js
```diff
@@ -70,6 +70,15 @@ suite('relative-time', function () {
     assert.equal(counter, 1)
   })
 
+  test('calls update even after nullish datetime', async () => {
```
cheshire137, line 73: Thank you for the quick fix! 🙇🏻‍♀️

PR #294, opened 2024-12-01
Overhaul duration calculation functions

leduyquang753:
The current implementation of the relative date display appears to be brittle and actually being broken, for instance with this PR's date being 2 Dec 2024, the date 28 Jan 2024 is being very wrongly displayed as "9 days ago".

image

This pull request is an attempt to rebuild the computations from the ground up so as to be more consistent and robust:

- The relative display format is now implemented using a separate function, taking into account the actual dates in question rather than relying on the information-stripped duration.
- The duration display format has a new implementation and no longer relies on a reference date.
- Functions `applyDuration` and `elapsedTime` now give consistent results.

Some semantics had to be changed and so I expect questions and probably some alterations to the details.

### DIFF
--- a/src/duration.ts
+++ b/src/duration.ts
@@ -1,6 +1,7 @@
 import DurationFormat from './duration-format-ponyfill.js'
 import type {DurationFormatOptions} from './duration-format-ponyfill.js'
 const durationRe = /^[-+]?P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?$/
+const durationRoundingThresholds = [Infinity, 11, 28, 21, 55, 55, 900]
 export const unitNames = ['year', 'month', 'week', 'day', 'hour', 'minute', 'second', 'millisecond'] as const
 export type Unit = typeof unitNames[number]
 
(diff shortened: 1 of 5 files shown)


src/duration.ts
```diff
@@ -86,134 +87,214 @@ export class Duration {
 
 export function applyDuration(date: Date | number, duration: Duration): Date {
   const r = new Date(date)
-  r.setFullYear(r.getFullYear() + duration.years)
-  r.setMonth(r.getMonth() + duration.months)
-  r.setDate(r.getDate() + duration.weeks * 7 + duration.days)
-  r.setHours(r.getHours() + duration.hours)
-  r.setMinutes(r.getMinutes() + duration.minutes)
-  r.setSeconds(r.getSeconds() + duration.seconds)
+  const actions: Array<() => void> = [
```
keithamus, line 90: I can't see why we need an array here. Because of the `if`? Having an array of functions and reversing them seems costly compared to duplicating the small number of lines twice in an `if/else`

src/duration.ts
```diff
@@ -86,134 +87,214 @@ export class Duration {
...
+  values.splice(2, 0, 0)
+  return new Duration(...values)
 }
 
-export function getRelativeTimeUnit(
-  duration: Duration,
-  opts?: Partial<RoundingOpts>,
-): [number, Intl.RelativeTimeFormatUnit] {
-  const rounded = roundToSingleUnit(duration, opts)
+export function getRoundedRelativeTimeUnit(rounded: Duration): [number, Intl.RelativeTimeFormatUnit] {
```
keithamus, line 297: If this is removing an unused variable that might be a great candidate for a smaller PR which can be approved and merged in isolation.

src/duration.ts
```diff
@@ -1,6 +1,7 @@
 import DurationFormat from './duration-format-ponyfill.js'
 import type {DurationFormatOptions} from './duration-format-ponyfill.js'
 const durationRe = /^[-+]?P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?$/
+const durationRoundingThresholds = [Infinity, 11, 28, 21, 55, 55, 900]
```
keithamus, line 4: I'm a little concerned that this array is a level of indirection which makes it hard to figure out the rounding rules. At minimum it might be worth adding comments to each of these explaining that they're months, days, hours and so-on.

keithamus (commented):
I've just left a couple of comments as jumping off points.

Overall I'm excited by this change but I am also a little concerned about how much is changing, especially with the tests. I'm in no way suggesting the current tests are especially correct but I also think we need to assess each one to see if the new value makes more sense, especially which when it comes to calendars (vs time) can be more of a colloquial preference than a hard science. For example we have found users tend to prefer the calendar year being used for such long durations, meaning Jan 2024-Sep 2022 is often rounded up to the much larger "2 years ago" - if we mark it as "1 year ago" we tend to get complaints :confused:.

leduyquang753:
> This PR seems too large to reasonably review and looking through some of the changes seem to not really change logic but just refactor for code style which makes it harder to review.

This is a complete rework of the logic, you can see the functions for elapsed time and rounding totally reimplemented with little resemblance to the original code, or you can try comparing outputs from the new implementation against the current one. The vast majority of changes are in the test suites, which I have also rebuilt to more comprehensively test the new behavior.

> Is it possible to break this PR down into smaller pieces and optimise for a clean diff?

I could split this into multiple parts if that is what you desire, namely the new elapsed time calculation, the new rounding function, the relative time implementation, and the improved duration application function. However adopting them one by one may lead to inconsistent behavior, unless the splitting is just for the purpose of more convenient code review.

keithamus:
> unless the splitting is just for the purpose of more convenient code review.

This is exactly the case. I am happy for us to release as one but it would be nice to get more detail _about the changes_ in a few separate PRs especially when it comes to the tests.

leduyquang753:
> For example we have found users tend to prefer the calendar year being used for such long durations, meaning Jan 2024-Sep 2022 is often rounded up to the much larger "2 years ago" - if we mark it as "1 year ago" we tend to get complaints 😕.

To specifically address this, my new implementation does output this duration as "in 2 years"/"2 years ago", the code only looks at the year difference in this case.

```
> new Intl.RelativeTimeFormat("en", {amounts: "auto"}).format(...relativeTime(new Date("2022-09-25"), "second", new Date("2024-01-15")))
'2 years ago'
```

> I also think we need to assess each one to see if the new value makes more sense

I completely understand this and so to make it clear I am not in any rush to get this new behavior merged right away, it will of course take some time to discuss all the little details.

leduyquang753:
> This is exactly the case. I am happy for us to release as one but it would be nice to get more detail *about the changes* in a few separate PRs especially when it comes to the tests.

Thank you for the confirmation. This will be changed to just a draft then and as I make the new set of PRs with more detailed descriptions it will be closed in favor of them.

src/duration.ts
leduyquang753: If the duration to be applied is negative, the order of the operations will be reversed. This is so that adding a duration then subtracting the same duration always gives the original date, which was not the case with the old implementation.

src/duration.ts
leduyquang753: As I have explained initially this is because the new implementation does not rely on a reference date to be passed anymore, but the original implementation did, and so this change can only come along with the new implementation of the function.

src/duration.ts
leduyquang753: I just attempted to place the constants at the same place that others appear to be placed, which is to the top of the source file. If putting them next to the function that uses them is better then I will happily do that, in fact I also do prefer it that way. I will be sure to add comments.

leduyquang753:
Closing in favor of split pull requests.

Issue #250, opened 2023-02-11
Wrong human text


antonio-fr:
I'm supposing this is the code that writes dates on Github web front.

1 year ago, it writes "2 years ago"
![image](https://user-images.githubusercontent.com/7551901/218257183-7b754dba-4614-4af8-a8a4-ce9a3bb9277e.png)

turbo:
Same issue on https://github.com/grayjoking/backhand (flagged up by user):

![image](https://user-images.githubusercontent.com/8023543/218694495-b1ab2700-c87b-458b-8d18-735e4f2832f0.png)

stapelberg:
Here’s a test case that illustrates the issue:

```diff
--- a/test/relative-time.js
+++ b/test/relative-time.js
@@ -2411,18 +2411,18 @@ suite('relative-time', function () {
         expected: 'in 3 years',
       },
       {
-        reference: '2022-12-31T12:00:00.000Z',
-        datetime: '2024-03-01T12:00:00.000Z',
-        tense: 'future',
-        format: 'micro',
-        expected: '3y',
+        reference: '2023-02-18T19:00:00.000Z',
+        datetime: '2021-02-02T20:49:31.000Z',
+        tense: 'past',
+        format: 'auto',
+        expected: '2 years ago',
       },
       {
-        reference: '2021-04-24T12:00:00.000Z',
-        datetime: '2023-02-01T12:00:00.000Z',
-        tense: 'future',
-        format: 'micro',
-        expected: '2y',
+        reference: '2023-02-18T19:00:00.000Z',
+        datetime: '2021-01-23T09:28:06.000Z',
+        tense: 'past',
+        format: 'auto',
+        expected: '2 years ago',
       },
     ])
```

Output:
```
test/relative-time.js:

 ❌ relative-time > legacy formats > <relative-time datetime="2021-02-02T20:49:31.000Z" tense="past" format="auto"  > => 2 years ago
      AssertionError: expected '3 years ago' to equal '2 years ago'
      + expected - actual
      
      -3 years ago
      +2 years ago
```

Might be related to https://github.com/github/relative-time-element/pull/249? cc @keithamus

keithamus:
I have a refactor PR that’ll be ready soon which will resolve these without regressing the other cases we’ve run into.

Wulfheart:
@keithamus currently I am trying to dive into the code. Could you please point me to the PR Seith the refactoring?

keithamus:
[...]

PR #296, opened 2024-12-04
Fix duration calculation

wxiaoguang:
Fix duration calculation

The simplest fix for https://github.com/github/relative-time-element/issues/250#issuecomment-2509670103 , not ideal, just a quick fix.


silverwind:
Maybe a test can be added here: 

https://github.com/github/relative-time-element/blob/e914da24dad0980cbfcb5dae6d56cf2c97659340/test/duration.ts#L231

wxiaoguang:
> Maybe a test can be added here:
> 
> https://github.com/github/relative-time-element/blob/e914da24dad0980cbfcb5dae6d56cf2c97659340/test/duration.ts#L231

If the project members have interests in this PR, I could add later.

francinelucca:
Thanks so much for diving into this @wxiaoguang! Looks like this proposed solution is causing one test to fail so we're unable to merge it at this time.
If you'd like to take a stab at fixing the tests you're welcome to do so, if not someone from the team will pick this up when we have the availability to do so.

You should be able to run the tests locally using `npm run test`

wxiaoguang:
> Thanks so much for diving into this @wxiaoguang! Looks like this proposed solution is causing one test to fail so we're unable to merge it at this time. If you'd like to take a stab at fixing the tests you're welcome to do so, if not someone from the team will pick this up when we have the availability to do so.
> 
> You should be able to run the tests locally using `npm run test`

But that failure seems not related. I have noticed that even on main branch, it still fails (that's another reason why I didn't touch the test code, I just would like to see whether it fails in your CI tasks).

````
 ❌ relative-time > [tense=future] > micro formats years
      AssertionError: expected '11y' to equal '10y'
      + expected - actual
      
      -11y
      +10y
      
      at n4.<anonymous> (test/relative-time.js:609:13)
````

**Comment by wxiaoguang** [comment]
The `11y vs 10y`  failure is caused by another bug in `elapsedTime`, `const month = Math.floor(day / 30)` is not right when the period is long enough.

File: src/duration.ts [hunk 1]
```diff
@@ -182,6 +182,9 @@ export function roundToSingleUnit(duration: Duration, {relativeTo = Date.now()}:
     } else if (monthsDiff < 11) {
       months = monthsDiff
       years = 0
+    } else if (monthsDiff === 11) {
+      months = 0
+      years += 1 // the old behavior: "11 months" is rounded to "1 year"
```
francinelucca, line 187: so in this case if the two dates were January and December, we'd round up to a year? wouldn't we want it to be 11 months?

francinelucca (commented):
Confirmed CI failure is present in prod, in that case I think we can move forward with this. @wxiaoguang Just have a logic question and adding tests I think is a great idea :)

src/duration.ts
wxiaoguang: > so in this case if the two dates were January and December, we'd round up to a year? wouldn't we want it to be 11 months?

The old code didn't do so due to ` if (monthsDiff < 11) { } else { year }` above

I have tried to make it support "11 months", it breaks a lot of tests.

wxiaoguang:
A test is added in fc858c3c34e038476008be60004b33ddfd368707 and added comments to clarify the problem.

wxiaoguang:
And made one more commit: if we'd like to support "11 months": 75800544934caeecdcd841dbe583ceb410a502e2 corrected the behavior and fixed the tests.

Feel free to choose the behavior you like.

wxiaoguang:
"Fixed" the "10y vs 11y" problem in da833cf59b93e06d91d314001b518d78219ec809, now the CI should be able to pass

viceice (approved):
🚀 ❤️

keithamus (approved):
This LGTM

francinelucca (approved):

francinelucca:
Merged! Thanks so much for doing this @wxiaoguang

PR #297, opened 2024-12-05
fix: wrap Intl.<>() calls in try/catch

francinelucca:
Closes https://github.com/github/primer/issues/4409

Fixes error thrown when an invalid locale is passed to the `Intl` functions by wrapping it in Try/Catch. When the call fails (locale is invalid), the default locale will be used

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -82,11 +82,13 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
   #updating: false | Promise<void> = false
 
   get #lang() {
-    return (
-      this.closest('[lang]')?.getAttribute('lang') ||
-      this.ownerDocument.documentElement.getAttribute('lang') ||
-      'default'
-    )
+    const lang = this.closest('[lang]')?.getAttribute('lang') ||
+      this.ownerDocument.documentElement.getAttribute('lang')
+    try {
+      return new Intl.Locale(lang ?? '').toString()
+    } catch {
+      return 'default'
+    }
   }
 
   #renderRoot: Node = this.shadowRoot ? this.shadowRoot : this.attachShadow ? this.attachShadow({mode: 'open'}) : this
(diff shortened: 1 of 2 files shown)


src/relative-time-element.ts
```diff
@@ -120,14 +120,27 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
...
+      dateTimeFormat = new Intl.DateTimeFormat('default', {
+        day: 'numeric',
+        month: 'short',
+        year: 'numeric',
+        hour: 'numeric',
+        minute: '2-digit',
+        timeZoneName: 'short',
+      })
+    }
+    return dateTimeFormat.format(date)
```
francinelucca, line 143: @keithamus I did some research on using an array in the constructor instead but it'd fail anyways if one of the locales was invalid, so I settled on this.
Also I'm not quite sure how to test this but happy to update on Primer and test once this is in

keithamus (commented):
Instead of try/catching each of the `Intl` calls we could add a try/catch into `#lang`:

```js
 get #lang() {
    const lang = this.closest('[lang]')?.getAttribute('lang') ||
      this.ownerDocument.documentElement.getAttribute('lang')
    try {
      return new Intl.Locale(lang).toString()
    } catch {
      return 'default'
    }
  }
```

This way `#lang` would always return a reliable language code and we minimise the surface area of changes.

francinelucca:
@keithamus done in [0a42e68](https://github.com/github/relative-time-element/pull/297/commits/0a42e68e491c1f68bdc0929c35e9725360460ec9), much better now, thanks!

keithamus (approved):
LGTM, some tests might be interesting. Also you’ll need someone from Primer to approve as I am no longer in the reviewers team 🥹

TylerJDev (approved):

keithamus (approved):
Overall LGTM, one nit:

PR #299, opened 2024-12-31
Reimplement `elapsedTime` to be precise

leduyquang753:
(This is part of a set of multiple pull requests looking to overhaul the calculation functions.)

The current implementation of `elapsedTime` just assumes there are 30 days in a month, which causes overshoots when `roundToSingleUnit` takes its result and, in a rather roundabout way, tries to compute the original date back.

This pull request reimplements the function so that it is consistent with `applyDuration`: applying the duration to the reference date always gives the original input date.

In math form:

```
If diff = date2 - date1 (elapsedTime)
then date2 = date1 + diff (applyDuration)
```

In code:

```js
applyDuration(date1, elapsedTime(date2, 'second', date1)) === date2
```

### DIFF
--- a/src/duration.ts
+++ b/src/duration.ts
@@ -95,27 +95,80 @@ export function applyDuration(date: Date | number, duration: Duration): Date {
   return r
 }
 
-export function elapsedTime(date: Date, precision: Unit = 'second', now = Date.now()): Duration {
-  const delta = date.getTime() - now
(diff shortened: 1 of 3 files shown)


francinelucca:
👋🏽 @leduyquang753 apologies for lack of review here. Is this still relevant? if so I'll be happy to review once conflicts are resolved

PR #303, opened 2025-01-13
Use node v22

camertron:
Should fix an issue encountered during release where the latest npm would not install in CI.


keithamus (approved):
LGTM

Issue #304, opened 2025-01-14
TypeScript types don't work with React 19


eramdam:
Today I ran into an issue where the types that are shipped with the packages do not work anymore when using React 19 in a TypeScript project. This is because the global `JSX` namespace has been removed. I managed to fix the issue by making this change using `patch-package` in my project.

```diff
diff --git a/node_modules/@github/relative-time-element/dist/relative-time-element-define.d.ts b/node_modules/@github/relative-time-element/dist/relative-time-element-define.d.ts
index 6465123..58113c8 100644
--- a/node_modules/@github/relative-time-element/dist/relative-time-element-define.d.ts
+++ b/node_modules/@github/relative-time-element/dist/relative-time-element-define.d.ts
@@ -9,6 +9,8 @@ declare global {
     interface HTMLElementTagNameMap {
         'relative-time': RelativeTimeElement;
     }
+}
+declare module 'react' {
     namespace JSX {
         interface IntrinsicElements {
             ['relative-time']: JSXBase['span'] & Partial<Omit<RelativeTimeElement, keyof HTMLElement>>;
```

keithamus:
Sounds like a worthwhile change. Do you think it’s backwards compatible or will we need to declare it twice?

eramdam:
I'm honestly not sure, the [React 19 upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript) makes it sound like it is _not_ backwards-compatible, but I didn't try declaring it twice either (on 18 and 19) 😅

eramdam:
@keithamus I've made a reproduction repo (https://github.com/eramdam/relative-element-ts-repro) and as far as I can tell, the fix seems to be backwards compatible with React 18.x using those dependencies:
```json
"dependencies": {
    "@github/relative-time-element": "^4.4.5",
    "react": "^18.3.1",
    "react-dom": "^18.3.1"
  },
  "devDependencies": {
    "@eslint/js": "^9.17.0",
    "@types/react": "^18.3.18",
    "@types/react-dom": "^18.3.5",
}
```

eramdam:
(sorry for the double notification, I messed up the config of the repo when I posted my first, now-deleted, comment) 

I made a [repro repository](https://github.com/eramdam/relative-element-ts-repro) and as far as I can tell, the fix is backwards compatible with React 18 / TypeScript 5.6 

```json
{
  "dependencies": {
[...]

PR #305, opened 2025-01-16
Declare `relative-element` under React.JSX to fix React 19 compatibility

eramdam:
Fixes https://github.com/github/relative-time-element/issues/304

I had to add a `@ts-expect-error` comment since otherwise the typechecker would complain about missing `react` as a dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,9 @@
         "eslint-plugin-custom-elements": "^0.0.8",
         "eslint-plugin-github": "^4.7.0",
         "typescript": "^5.0.4"
+      },
+      "peerDependencies": {
+        "@types/react": "18 || 19"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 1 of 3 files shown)


keithamus:
@github/primer you'll need to integrate this to check whether or not this is viable for the monolith. Please do not merge before checking this, otherwise we'll have to revert.

keithamus:
@eramdam looks like this isn't quite passing the CI checks. Does `tsc --noEmit` pass locally for you?

eramdam:
> @eramdam looks like this isn't quite passing the CI checks. Does `tsc --noEmit` pass locally for you?

Oops, you're right, it doesn't 🙈  I wanted to avoid it but it looks like the only clean way of fixing the issue was to add `@types/react` as a peer dependency so I did that.

camertron:
Hey @eramdam, thanks for looking into this. I'm hesitant to merge React-specific code into this project, considering it really has nothing to do with React. Could we instead put the React code into a separate .d.ts file and ask that clients add it to their tsconfig.json?

keithamus (approved):
I think this looks okay from a library perspective. I dislike that there’s react specific idioms too, but react is always messy to integrate with so… 🤷‍♂️

@camertron lets try integrating this with PVC and see if there are any issues. AIUI this shouldn’t cause too much friction within PVC, but it might and at that point we’ll need to look at other solutions. My hope is this is a transparent change to non-react projects like PVC, and doesn’t pull in additional dependencies.

alanbernale:
It’d be really helpful to merge this pull request to fix compatibility with React 19.

lesliecdubs:
> @camertron lets try integrating this with PVC and see if there are any issues. AIUI this shouldn’t cause too much friction within PVC, but it might and at that point we’ll need to look at other solutions. My hope is this is a transparent change to non-react projects like PVC, and doesn’t pull in additional dependencies.

@hectahertz as our PVC release conductor this week, could you take a look at integrating this change with PVC and see if there are any issues?

francinelucca:
FYI, I asked @jonrohan about this last week:

J: for the releative-time-element, I think we just need to test upgrading in PVC and see if it still runs properly on the lookbook
M: won't we need to publish first anyways? :thinkin: I don't see canaries being deployed
J: hmm yeah I’m not sure we have that setup there :disappointed:
M: merge -> release -> test in PVC -> (hopefully not) rollback, then?
J: 👍 

So the work here is:
- Merge and Release this
- Upgrade PVC to this release version and not any errors
- If no errors, this is done 
- If errors, we'll need to roll this back

francinelucca (approved):

PR #312, opened 2025-04-25
test: update assertion for test

joshblack:
Noticed this was failing in newer PRs so updated it to pass. Not sure what the correct value is here but `10y` seemed to make sense because the test was subtracting 10 years.


francinelucca (approved):
😅

PR #315, opened 2025-05-07
fix: Ensure aria-hidden applies to shadow root

khiga8:
Relates to: https://github.com/github/pull-requests/issues/17192

## Problem

When `aria-hidden="true"` is set on an relative-time element, the attribute doesn't always persist into the shadow root. As a result, even when `aria-hidden` is applied, the text rendered in relative-time can end up composing the accessible name (instead of being hidden from the AT tree).

## Solution

This makes update to render the text content within `aria-hidden="true"`, when `aria-hidden="true"` is set on the shadow root. 

## Demo

Before (no sound):

https://github.com/user-attachments/assets/5a049c15-8ec7-4a7f-be6e-97c3b149f389

video description: I'm on the demo page with the developer console open. The markup shows `aria-hidden` being toggled between true and false relative-time upon pressing the button. I switch to the accessibility tree mode. I see the accessible name for the button remains: `With aria-hidden on Dec 31, 1969` despite toggling.

After (no sound):

https://github.com/user-attachments/assets/b662b53c-11c2-49b4-98b5-8bf002b04b0f

video description: I'm on the demo page with the developer console open. The markup shows `aria-hidden` being toggled between true and false relative-time upon pressing the button. I switch to the accessibility tree mode. I see the accessible name for the button toggle between: `With aria-hidden on Dec 31, 1969` and `With aria-hidden` as we'd expect.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -200,6 +200,8 @@
     </relative-time>
   </p>
 
+  <h2>With Aria Hidden</h2>
+
   <button class="js-toggle-aria-hidden">
     With aria-hidden
     <relative-time aria-hidden="true" datetime="1970-01-01T00:00:00.000Z">
--- a/examples/index.html
+++ b/examples/index.html
@@ -200,17 +200,17 @@
     </relative-time>
   </p>
 
-  <p>
+
+  <button class="js-toggle-aria-hidden">
     With aria-hidden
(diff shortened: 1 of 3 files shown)


JoyceZhu (approved):
🚀

jonrohan (approved):
Nice catch

PR #316, opened 2025-06-20
feat: add support for time-zone attribute

### LINKED ISSUE #292: Allow setting timezone attribute
Have a case where sometimes we want to display times in local TZ, sometimes based on TZ of where the timestamp originated. I've just slapped a TZ attribute on the element and passed it through - is there particular reason this isn't in the API already before I try to upstream it? Did I fail to RTFM?

davhdavh:
fixes #292


TylerJDev:
Hey @davhdavh! 👋 

Thanks so much for the PR! We'll aim to take a look at it and review it this week! 😁

francinelucca:
@TylerJDev is this good to merge?

PR #323, opened 2025-09-18
Change 'just now' to 'now' in README

### LINKED ISSUE #322: Readme incorrect says that the date will be formatted as 'just now'
The library changed many moons ago to remove the 'just', specifically in #104 however, the README at the root of the repo still says that dates will be formatted as `just now`.

darthsteven:
Fixes #322


francinelucca (approved):
Thanks for the fix!

PR #330, opened 2025-10-28
Experimental: `<relative-time>` supports thresholds that are specified by the user

zaataylor:
These changes enable the `<relative-time>` element to support user-supplied thresholds that pull from the value of a newly defined `data-preferred-threshold` attribute. The distinction between this feature and the pre-existing `threshold` support offered by the element already is that it will theoretically enable per-user specification of preferred thresholds over all `<relative-time>` elements. This would enable us to support something akin to a user setting allowing a developer to decide how long times show up as relative before they are shown as absolute time representations.


zaataylor:
closing for now

PR #331, opened 2025-11-06
Experimental: Check for override attribute `disable-preference-override`

zaataylor:
Now, part of the check for determining if we should display user time preferences will also check for whether there is a specific attribute on the element, `disable-preference-override`. If this attribute is present, we will use whatever the default behavior/display options are for the element, rather than displaying with absolute time.

This will enable us to selectively override elements in certain UIs even when a user setting to enable absolute time has been enabled. This will allow us to roll out changes without breaking parts of the UI we already know can't handle longer timestamps well.

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -245,7 +245,7 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     if (format === 'duration') return false
 
     // Override for user preferences; used sparingly to preserve intended relative formatting in some places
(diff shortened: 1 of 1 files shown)


llastflowers (approved):

khiga8 (commented):
Looking good! Can we add a test? Maybe somewhere in [this suite](https://github.com/github/relative-time-element/blob/main/test/relative-time.js#L1888)!

zaataylor:
closing for now, we can revive this later if needed

Issue #353, opened 2026-06-02
`format="micro"` ignores `threshold` and has no `tense` support  


maxbeizer:
### Problem

We want compact relative timestamps on GitHub Issues — **"2w ago"** for recent items, falling back to absolute dates (**"Jan 1, 2024"**) for items older than 30 days. The natural API for this would be:

```html
<relative-time format="micro" tense="past" threshold="P30D" datetime="2024-01-01T00:00:00Z" />
```

But today this doesn't work because there are two related gaps in `format="micro"`:

1. **`threshold` is ignored for `format="micro"`** — `resolveFormat()` returns `'duration'` unconditionally for micro, never checking whether an explicitly configured threshold should switch the element to absolute datetime output. Old dates therefore render as micro durations like `"6mo"` or `"2yr"`.

2. **`micro` output cannot include tense phrasing** — `getDurationFormat()` returns only the compact duration string, so consumers cannot get `"2w ago"` or `"in 3d"` from the component alone.

### Current workaround (and why it's bad)

Consumers are forced to add a separate suffix element outside the component:

```tsx
const MODULE_LOAD_TIME = Date.now()
const THIRTY_DAYS_MS = 30 * 86_400_000

// in render:
<RelativeTime format="micro" datetime={createdAt} threshold="P30D" />
{new Date(createdAt).getTime() > MODULE_LOAD_TIME - THIRTY_DAYS_MS && <span> ago</span>}
```

This has several problems:
- **Desync**: The web component live-updates on a timer with `Date.now()`, but the external suffix is keyed to a stale timestamp captured at module load. At the threshold boundary, they disagree — producing either `"Jan 1, 2024 ago"` or `"4w"` (no ago).
- **User preference**: Users who set `data-prefers-absolute-time="true"` get the component's absolute date plus the hardcoded suffix: `"Jan 01, 2020, 08:02:03 PM ago"`.
- **DRY**: The workaround has to be duplicated in every component that uses compact timestamps.
- **Threshold doesn't actually work**: As noted above, `threshold="P30D"` is silently ignored for micro, so old items show `"6mo"` instead of an absolute date.

#### 1. `format="micro"` should support explicit `threshold`

When a `threshold` attribute is explicitly provided and the elapsed duration exceeds that threshold, `format="micro"` should fall back to datetime format (absolute date). This gives consumers an opt-in API for compact recent timestamps with absolute old timestamps:

```html
<relative-time format="micro" tense="past" threshold="P30D" datetime="..." />
```

[...]

PR #354, opened 2026-06-02
Support explicit micro thresholds

maxbeizer:
## Summary

`format="micro"` currently ignores `threshold` because it always resolves to duration output. This makes consumers compose their own cutoff logic, which can drift from the element's live updates.

This PR makes threshold support opt-in for micro timestamps:
- `format="micro" threshold="P30D"` shows compact durations within the threshold and absolute dates outside it
- existing `format="micro"` usages without `threshold` keep rendering old dates as compact durations
- threshold comparisons use full elapsed time, independent of display precision
- future dates outside the threshold schedule an update at the threshold boundary so they switch to micro output on time
- `data-prefers-absolute-time="true"` now applies to micro output while `duration` and `elapsed` remain excluded

Refs #353

## Tests

- `npm test`
- `npm run lint`

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,9 @@ This is identical to `format=relative`. Code that uses `format=auto` should migr
 
 ###### `format=micro`
 
-The `micro` format which will display relative dates (within the threshold) in a more compact format. Similar to `relative`, the `micro` format rounds values to the nearest largest value. Additionally, `micro` format will not round _lower_ than 1 minute, as such a `datetime` which is less than a minute from the current wall clock time will display `'1m'`.
+The `micro` format displays relative dates in a more compact format. Similar to `relative`, the `micro` format rounds values to the nearest largest value. Additionally, `micro` format will not round _lower_ than 1 minute, as such a `datetime` which is less than a minute from the current wall clock time will display `'1m'`.
+
+If the `threshold` attribute is explicitly set, `micro` will display compact relative dates within the threshold and absolute dates outside of it. If `threshold` is not set, `micro` will continue to display compact relative dates without applying the default threshold.
 
 Code that uses `format=micro` should consider migrating to `format=relative` (perhaps with `formatStyle=narrow`), as `format=micro` can be difficult for users to understand, and can cause issues with assistive technologies. For example some screen readers (such as VoiceOver for mac) will read out `1m` as `1 meter`.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -230,11 +232,11 @@ Precision can be used to limit the display of an `relative` or `duration` format
 
 ##### threshold (`string`, default: `P30D`)
 
-If `tense` is anything other than `'auto'`, or `format` is `'relative'` (or the deprecated `'auto'` or `'micro'` values), then this value will be ignored.
+If `tense` is anything other than `'auto'`, or `format` is `'relative'` (or the deprecated `'auto'` value), then this value will be ignored. For `format="micro"`, `threshold` is honored only when the `threshold` attribute is explicitly set.
```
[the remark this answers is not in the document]
maxbeizer: Addressed in 26ca974 by correcting the threshold docs: threshold applies to relative/auto when tense is auto, and micro only honors it when explicitly set.

src/relative-time-element.ts
```diff
@@ -305,8 +308,8 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
   }
 
   #shouldDisplayUserPreferredAbsoluteTime(format: ResolvedFormat): boolean {
-    // Never override duration format with absolute format.
-    if (format === 'duration') return false
+    // Never override duration or elapsed format with absolute format.
+    if (format === 'duration' && this.format !== 'micro') return false
```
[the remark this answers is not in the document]
maxbeizer: Addressed in 26ca974 by making observation conditional on rendered output: user-preferred absolute micro output no longer keeps duration-style polling except for Today/current-year absolute-time updates.

src/relative-time-element.ts
```diff
@@ -26,10 +37,40 @@ function getUnitFactor(el: RelativeTimeElement): number {
...
+  let factor = 60 * 60 * 1000
+  if (ms < 60 * 1000) {
+    factor = 1000
+  } else if (ms < 60 * 60 * 1000) {
+    factor = 60 * 1000
+  }
+  const threshold = getExplicitThreshold(el)
+  if (el.format === 'micro' && threshold) {
+    const thresholdDuration = Duration.from(threshold)
+    const signedThresholdDuration = date.getTime() > now ? negateDuration(thresholdDuration) : thresholdDuration
```
maxbeizer, line 50: This is the important live-update bit for threshold fallback: the normal observer cadence can be too late, so explicit micro thresholds schedule at `date - threshold` for future dates and `date + threshold` for past dates.

src/relative-time-element.ts
```diff
@@ -45,15 +86,16 @@ function isBrowser12hCycle(): boolean {
 const dateObserver = new (class {
   elements: Set<RelativeTimeElement> = new Set()
   time = Infinity
+  updating = false
 
   observe(element: RelativeTimeElement) {
-    if (this.elements.has(element)) return
     this.elements.add(element)
+    if (this.updating) return
```
maxbeizer, line 93: This intentionally allows already-observed elements to be rescheduled when `threshold` is added or changed. The `updating` guard keeps that from creating duplicate timers while the observer loop is running.

src/relative-time-element.ts
```diff
@@ -561,7 +615,7 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     }
 
     const duration = elapsedTime(date, this.precision, now)
-    const format = this.#resolveFormat(duration)
+    const format = this.#resolveFormat(duration, elapsedTime(date, 'millisecond', now))
```
maxbeizer, line 618: Using a millisecond-precision duration here keeps threshold resolution independent of display precision. Micro still renders with its normal minute floor, but threshold cutoffs are based on the actual elapsed time.

src/relative-time-element.ts
```diff
@@ -590,8 +644,8 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     }
 
     const shouldObserve =
-      format === 'relative' ||
-      format === 'duration' ||
+      (!displayUserPreferredAbsoluteTime && (format === 'relative' || format === 'duration')) ||
```
maxbeizer, line 647: This keeps user-preferred absolute output from polling like duration output. Micro only remains observed for explicit threshold transitions or the existing Today/current-year absolute-time cases.

test/relative-time.js
```diff
@@ -40,6 +40,44 @@ suite('relative-time', function () {
...
+    }
+    try {
+      time.setAttribute('format', 'micro')
+      time.setAttribute('datetime', new Date(Date.now() + 65 * 1000).toISOString())
+      await Promise.resolve()
+      assert.equal(time.shadowRoot.textContent, '1m')
+      assert.isAtLeast(delays[0], 59000)
+
+      delays.length = 0
+      time.setAttribute('threshold', 'PT1M')
```
maxbeizer, line 59: This test covers the dynamic attribute case: a micro element can already be observed, then gain an explicit threshold that requires an earlier reschedule.

liuliu-dev (approved):
LGTM! ❤️

maxbeizer:
@liuliu-dev as a first time contributor to this repo, do you merge it? Do you cut a new version? Who bumps dotcom? Happy to help just don't want to overstep

PR #355, opened 2026-06-02
Add micro tense phrasing

### LINKED ISSUE #353: `format="micro"` ignores `threshold` and has no `tense` support

maxbeizer:
## Summary

This builds on #354 by making `format="micro"` useful with explicit tense. Instead of forcing consumers to append their own `ago` suffix or `in` prefix, micro output now uses localized `Intl.RelativeTimeFormat` narrow phrasing.

Behavior after this PR:
- `format="micro" tense="past"` renders values like `2w ago`
- `format="micro" tense="future"` renders values like `in 3d`
- `tense="auto"` keeps the existing compact output like `2w`
- datetime fallback and user-preferred absolute-time output are not wrapped with tense text

Closes #353

## Tests

- `npm test`
- `npm run lint`

### DIFF
--- a/src/relative-time-element.ts
+++ b/src/relative-time-element.ts
@@ -236,6 +236,28 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     return 'datetime'
   }
 
+  #getMicroDuration(duration: Duration): Duration {
+    duration = roundToSingleUnit(duration)
+    // Allow month-level durations to pass through even with mismatched tense
+    if (
+      duration.months === 0 &&
+      ((this.tense === 'past' && duration.sign !== -1) || (this.tense === 'future' && duration.sign !== 1))
+    ) {
+      return microEmptyDuration
+    }
+    return duration
+  }
+
+  #getMicroRelativeFormat(duration: Duration): string {
+    const relativeFormat = new Intl.RelativeTimeFormat(this.#lang, {
+      numeric: 'always',
+      style: 'narrow',
+    })
+    duration = this.#getMicroDuration(duration)
+    const [int, unit] = getRelativeTimeUnit(duration.blank ? microEmptyDuration : duration)
+    return relativeFormat.format(Math.abs(int) * (this.tense === 'past' ? -1 : 1), unit)
+  }
+
   #getDurationFormat(duration: Duration): string {
     const locale = this.#lang
     const format = this.format
@@ -243,15 +265,8 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
     const tense = this.tense
     let empty = emptyDuration
     if (format === 'micro') {
-      duration = roundToSingleUnit(duration)
+      duration = this.#getMicroDuration(duration)
       empty = microEmptyDuration
-      // Allow month-level durations to pass through even with mismatched tense
-      if (
-        duration.months === 0 &&
-        ((this.tense === 'past' && duration.sign !== -1) || (this.tense === 'future' && duration.sign !== 1))
-      ) {
-        duration = microEmptyDuration
-      }
     } else if ((tense === 'past' && duration.sign !== -1) || (tense === 'future' && duration.sign !== 1)) {
       duration = empty
     }
@@ -624,7 +639,11 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
       newText = this.#getUserPreferredAbsoluteTimeFormat(date)
     } else {
       if (format === 'duration') {
-        newText = this.#getDurationFormat(duration)
+        if (this.format === 'micro' && this.tense !== 'auto' && Intl.RelativeTimeFormat) {
+          newText = this.#getMicroRelativeFormat(duration)
+        } else {
+          newText = this.#getDurationFormat(duration)
+        }
       } else if (format === 'relative') {
         newText = this.#getRelativeFormat(duration)
       } else {
--- a/test/relative-time.js
+++ b/test/relative-time.js
@@ -740,7 +740,7 @@ suite('relative-time', function () {
       assert.equal(time.shadowRoot.textContent, 'in 2y')
     })
 
-    test('micro formats past times', async () => {
+    test('micro formats near-future times', async () => {
       const now = new Date(Date.now() + 3 * 1000).toISOString()
       const time = document.createElement('relative-time')
       time.setAttribute('tense', 'future')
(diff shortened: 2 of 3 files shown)


src/relative-time-element.ts
```diff
@@ -236,22 +236,37 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
...
+      duration.months === 0 &&
+      ((this.tense === 'past' && duration.sign !== -1) || (this.tense === 'future' && duration.sign !== 1))
+    ) {
+      return microEmptyDuration
+    }
+    return duration
+  }
+
+  #getMicroRelativeFormat(duration: Duration): string {
+    const relativeFormat = new Intl.RelativeTimeFormat(this.#lang, {
```
maxbeizer, line 252: This uses `Intl.RelativeTimeFormat` rather than string-concatenating English `ago` / `in`, so the compact tense output still follows `lang`.

src/relative-time-element.ts
```diff
@@ -624,7 +639,11 @@ export class RelativeTimeElement extends HTMLElement implements Intl.DateTimeFor
       newText = this.#getUserPreferredAbsoluteTimeFormat(date)
     } else {
       if (format === 'duration') {
-        newText = this.#getDurationFormat(duration)
+        if (this.format === 'micro' && this.tense !== 'auto' && Intl.RelativeTimeFormat) {
```
maxbeizer, line 642: The tense phrasing is intentionally limited to micro duration output. Datetime fallback from #354 and user-preferred absolute output bypass this path, so we do not produce strings like `Jan 1, 2024 ago`.

test/relative-time.js
```diff
@@ -504,6 +527,17 @@ suite('relative-time', function () {
       await Promise.resolve()
       assert.equal(time.shadowRoot.textContent, 'hace 3 días')
     })
+
+    test('micro tense phrasing respects lang attribute', async () => {
```
maxbeizer, line 531: This is the guard against accidentally making the feature English-only: the micro tense path should localize through `Intl.RelativeTimeFormat`.

liuliu-dev (approved):

PR #368, opened 2026-07-29
Fix year precision for duration and elapsed formats

### LINKED ISSUE #366: precision="year" on duration/elapsed format is ignored — falls back to full millisecond precision
## Version
5.3.0

## Affected file
`src/duration.ts` (compiled: `dist/duration.js`), function `elapsedTime()`

## Description
Setting `precision="year"` on a `<relative-time>` (or `<time-ago>`) element with `format="duration"` (or `format="elapsed"`) does not limit the output to years — instead it renders the full duration breakdown down to milliseconds, as if no `precision` had been set at all.

## Reproduction
```html
<relative-time datetime="2023-06-23T08:38:34.578Z" format="duration" precision="year">
</relative-time>
```

Expected: `3 years`

Actual: `3 years, 1 month, 21 days, 10 minutes, 56 seconds, 181 milliseconds`

Every other precision value (`month`, `day`, `hour`, `minute`, `second`) works correctly and truncates the output as documented. 

## Root cause
In `elapsedTime()`:

```js
export function elapsedTime(date, precision = 'second', now = Date.now()) {
  // ...
  const i = unitNames.indexOf(precision) || unitNames.length;
  // ...
}
```

`unitNames` is `['year', 'month', 'week', 'day', 'hour', 'minute', 'second', 'millisecond']`.

`unitNames.indexOf('year')` returns `0`. Since `0` is falsy in JavaScript, the `||` operator treats it as "not found" and falls back to `unitNames.length` (`8`), which effectively means "no truncation — include every unit down to milliseconds."

## Suggested fix
Use a strict check for "not found" instead of relying on truthiness:

```js
const idx = unitNames.indexOf(precision);
const i = idx === -1 ? unitNames.length : idx;
```

Pybsama:
## Summary

- honor `precision="year"` when calculating elapsed durations
- preserve the existing empty-duration result for an unknown runtime precision
- cover positive and negative helper results plus rendered `duration` and `elapsed` output

## Root cause

`unitNames.indexOf("year")` returns `0`, but the logical-or expression treated
that valid index as falsy and replaced it with `unitNames.length`. Using the raw
index keeps `0` as the year cutoff and also preserves the existing `-1`
behavior for an unknown runtime value.

## Testing

- `npm test` (634 passing)
- `npm run lint`
- `npm run build`

Fixes #366

### DIFF
--- a/src/duration.ts
+++ b/src/duration.ts
@@ -156,7 +156,8 @@ export function elapsedTime(date: Date, precision: Unit = 'second', now = Date.n
   const day = Math.floor(hr / 24)
   const month = Math.floor(day / 30)
   const year = Math.floor(month / 12)
-  const i = unitNames.indexOf(precision) || unitNames.length
+  const precisionIndex = unitNames.indexOf(precision)
+  const i = precisionIndex === -1 ? unitNames.length : precisionIndex
   return new Duration(
     i >= 0 ? year * sign : 0,
     i >= 1 ? (month - year * 12) * sign : 0,
--- a/src/duration.ts
+++ b/src/duration.ts
@@ -156,8 +156,7 @@ export function elapsedTime(date: Date, precision: Unit = 'second', now = Date.n
   const day = Math.floor(hr / 24)
   const month = Math.floor(day / 30)
   const year = Math.floor(month / 12)
-  const precisionIndex = unitNames.indexOf(precision)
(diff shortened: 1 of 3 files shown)


liuliu-dev (approved):
Thanks for the fix! ❤️

PR #378, opened 2026-09-07
Warn for non-standard datetime values

### LINKED ISSUE #266: Stricter `datetime` validation
The docs about `datetime` argument say:

> This must be a valid [ISO8601 DateTime](https://en.wikipedia.org/wiki/ISO_8601)

This statement is incorrect. Any value that can be passed to the [`Date` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format) will work which present a cross-browser issue because JS engines are inconsistent in which formats their `Date` constructor will accept.

For example, the golang default string representation only parses in v8:

```bash
$ eshost -e "new Date('2009-11-10 23:00:00+00:00 UTC')"
#### JavaScriptCore
Invalid Date

#### spidermonkey
Invalid Date

#### v8
Wed Nov 11 2009 00:00:00 GMT+0100 (Central European Standard Time)
```

How about adding an option attribute to pass in a validation regex into the element to validate the passed dates, when present? This would at least not make this issue missable by developers who only test in Chrome.

ryux1:
## Summary

- validate non-empty `datetime` attributes against the ECMAScript date-time string format
- warn for malformed or implementation-dependent values while preserving the existing parsing behavior
- cover valid date-only, date-time, offset, expanded-year, and midnight forms

Closes #266

## Testing

- `npm test` (636 Chromium tests)
- TypeScript and ESLint checks run as part of the test build
- `npx prettier --check src/relative-time-element.ts test/relative-time.js`
- `git diff --check`


francinelucca (commented):
Thanks for this contribution @ryux1 🙏🏽, can you take a look at copilot's review comments above? thanks!

